PR #28, opened 2020-11-03
Update to mmtk-core PR #161

qinsoon:


javadamiri (approved):
LGTM (update rev before merge)

PR #42, opened 2021-01-18
Run-time plan selection

qinsoon:


openjdk/mmtkBarrierSet.hpp
```diff
@@ -99,49 +95,49 @@ class MMTkBarrierSet : public BarrierSet {
     template <typename T>
     static void oop_store_in_heap(T* addr, oop value) {
       Raw::oop_store(addr, value);
-#if MMTK_ENABLE_WRITE_BARRIER
-      MMTkBarrierRuntime::record_modified_edge((void*) addr);
-#endif
+      if (MMTkBarrierSet::enable_write_barrier) {
```
wenyuzhao, line 98: I'll figure out a way to select between multiple different `BarrierSet`s instead of branching here. Probably in another PR.

wenyuzhao (approved):
Looks good to me

openjdk/mmtkBarrierSet.hpp
qinsoon: Is this code used in C1/C2 JIT compiler or the interpreter?

PR #44, opened 2021-02-04
Select barrier based on active plan

wenyuzhao:
This PR refactors the barrier related code, makes it easier to add a new barrier or modify an existing one.

The activated barrier depends on the runtime-selected plan.

---

# Runtime barrier selection

Old approach of the `openjdk_needs_write_barrier()` global flag can only enable or disable the use of write barriers. This works because we currently only have one barrier implementation available -- the object remembering barrier. But we want to support more barrier algorithms in the future.

This PR adds a new interface `char* mmtk_active_barrier()` to make VM be able to choose and enable one barrier among all the available barrier implementations.

# Refactor

Definition and code for different barriers are move to different cpp files. Dynamic-dispatch is used to indirect the barrier calls to the selected barrier implementation.

# Performance

The dynamic-dispatch based indirection can cause performance drops, especially the runtime part -- We now have a function call for each cpp-implemented barrier invocation (invoked by the vm code).

JIT code should be fine... the old implementation uses dynamic-dispatch anyway.

# TODO

- [x] Choose barriers based on the selected plan's configuration
- [ ] Benchmark


openjdk/mmtkBarrierSet.hpp
```diff
@@ -87,57 +116,44 @@ class MMTkBarrierSet : public BarrierSet {
...
   private:
     typedef BarrierSet::AccessBarrier<decorators, BarrierSetT> Raw;
   public:
     template <typename T>
     static void oop_store_in_heap(T* addr, oop value) {
-      Raw::oop_store(addr, value);
-      if (MMTkBarrierSet::enable_write_barrier) {
-        MMTkBarrierRuntime::record_modified_edge((void*) addr);
-      }
+      UNREACHABLE();
```
qinsoon, line 128: What's the reason that some of the methods are unreachable now?

qinsoon (approved):
Looks good to me.

openjdk/mmtkBarrierSet.hpp
wenyuzhao: Because these methods are never called by hotspot (at lease when running dacapo benchmarks). Probably they are just legacy interfaces...

PR #46, opened 2021-02-11
MallocMS

paigereeves:
This PR includes changes to the OpenJDK binding to support MallocMS. Related PR to mmtk-core is [#235](https://github.com/mmtk/mmtk-core/pull/235)


qinsoon:
I'll push a fix for the CI issue.

qinsoon:
@paigereeves Can you merge with master? The issue should be fixed with #47.

paigereeves:
@qinsoon Do you know why this error is appearing? I don't think I've seen it before.

qinsoon:
> @qinsoon Do you know why this error is appearing? I don't think I've seen it before.

I haven't seen this before. Can you check the following?
1. Revert the local path changes in `Cargo.toml` (though it is comments, the CI does use it).
2. Is there any change to the OpenJDK fork as you updated the `openjdk` submodule to a new revision in the PR?. Unless you have some changes to the VM (you shouldn't need to though), you can revert it back.

paigereeves:
I did need to make a change to the VM at [collectedHeap.hpp](https://github.com/mmtk/openjdk/blob/master/src/hotspot/share/gc/shared/collectedHeap.hpp), making the function `is_in_reserved` virtual so that it accounts for malloc allocating outside the MMTk heap.

I've tried resetting to upstream, yet I'm still getting this same error when I try to configure.

qinsoon:
> I did need to make a change to the VM at [collectedHeap.hpp](https://github.com/mmtk/openjdk/blob/master/src/hotspot/share/gc/shared/collectedHeap.hpp), making the function `is_in_reserved` virtual so that it accounts for malloc allocating outside the MMTk heap.

Is there any chance to refactor so we don't need to modify the VM? You can declare that function in the binding I think.

> I've tried resetting to upstream, yet I'm still getting this same error when I try to configure.

Did you try resetting to upstream locally? If so, that means it is not a CI issue.

javadamiri:
> @qinsoon Do you know why this error is appearing? I don't think I've seen it before.

mmtk/Cargo.toml
```diff
@@ -23,7 +23,7 @@ lazy_static = "1.1"
 # But other changes including adding/removing whitespaces in commented lines may break the CI.
 mmtk = { git = "ssh://git@github.com/mmtk/mmtk-core.git", rev = "b721a6d29d736282c36aa01196efd8f5bf26567f" }
 # Uncomment the following to build locally
-# mmtk = { path = "../repos/mmtk-core" }
+# mmtk = { path = "../../mmtk-core" }
```
wenyuzhao, line 26: Please revert this line. As mmtk-core sits at `../repos/mmtk-core` by default.

openjdk/mmtkBarrierSet.hpp
```diff
@@ -35,7 +35,7 @@
 #include "utilities/fakeRttiSupport.hpp"
 #include "mmtk.h"
 
-#define MMTK_ENABLE_ALLOCATION_FASTPATH true
+#define MMTK_ENABLE_ALLOCATION_FASTPATH false
```
wenyuzhao, line 38: Turn-off the fast path will also slow down the other collectors (by a lot). We need to figure out a way to only disable fast path for your mallocms gc. Probably we need to refactor this code.

openjdk/mmtkHeap.cpp
```diff
@@ -164,7 +164,7 @@ bool MMTkHeap::is_in(const void* p) const {
    //return cp >= committed_low_addr() && cp < committed_high_addr();
 
    //guarantee(false, "is in not supported");
-    return p >= (void*) _start && p < (void*) _end;
+    return p >= (void*) _start && p < (void*) _end || is_in_reserved_malloc(p) == 1;
```
wenyuzhao, line 167: For MallocMS this is necessary. But for other GCs, calling into rust code every time at here may slow down the performance.

I'm not familiar with @qinsoon's change but I think we have a global variable in the cpp side to record the selected collector. We can add a branch here and call `is_in_reserved_malloc` only if MallocMS is the selected collector.

openjdk/mmtkHeap.cpp
```diff
@@ -179,7 +179,7 @@ bool MMTkHeap::is_in_reserved(const void* p) const {
     //guarantee(false, "is in not supported");
 
     //printf("calling MMTkHeap::is_in_reserved\n");
-    return p >= (void*) _start && p < (void*) _end;
+    return p >= (void*) _start && p < (void*) _end || is_in_reserved_malloc(p) == 1;
```
wenyuzhao, line 182: Same here

openjdk/mmtkMutator.hpp
```diff
@@ -38,9 +39,18 @@ struct LargeObjectAllocator {
     RustDynPtr plan;
 };
 
+struct MallocAllocator {
+    void* tls;
+    void* cursor;
+    void* limit;
+    RustDynPtr space;
+    void* plan;
+};
```
wenyuzhao, line 48: This is inconsist with the MallocAllocator definition in Rust:

```rust
#[repr(C)]
pub struct MallocAllocator<VM: VMBinding> {
    pub tls: OpaquePointer,
    space: Option<&'static dyn Space<VM>>,
    plan: &'static dyn Plan<VM = VM>,
}
```

So here we only need three c++ struct fields, the first is a normal pointer, and the other two are rust dyn pointers:

```c++
struct MallocAllocator {
    void* tls;
    RustDynPtr space;
    RustDynPtr plan;
};
```

wenyuzhao (changes requested):

openjdk/mmtkBarrierSet.hpp
wenyuzhao: I think `MMTK_ENABLE_ALLOCATION_FASTPATH` does not have to be a macro. We can turn this into a global variable (or probably a static field in MMTkHeap) and during VM-initialisation, set this to false only if MallocMS is the selected collector.

openjdk/mmtkHeap.cpp
wenyuzhao: And I suggest we put an assertion (debug_assert) in `is_in_reserved_malloc` function body to ensure the selected collector is MallocMS. This prevents other collectors from calling this function.

openjdk/mmtkHeap.cpp
wenyuzhao: Looks like the `is_in_reserved` and `is_in` are duplications. You can just indirect one to the other to reduce the duplication, instead of copying the code twice.

javadamiri: These two functions are present in the master branch as well.
They are not added by Paige.
Anyway, I'll redirect one to the other for now.

javadamiri:
> > I did need to make a change to the VM at [collectedHeap.hpp](https://github.com/mmtk/openjdk/blob/master/src/hotspot/share/gc/shared/collectedHeap.hpp), making the function `is_in_reserved` virtual so that it accounts for malloc allocating outside the MMTk heap.
> 
> Is there any chance to refactor so we don't need to modify the VM? You can declare that function in the binding I think.

I don't think there is any way to make it work without this modification.
I think the reason why it's been working so far is that the implementation of `is_in_reserved` in `collectedHeap` was the same as its implementation in `mmtkHeap`. So, although the overriden implementation of the function was never called, the stock implementation was producing the same result. (They both checked verses a fixed `start` and `end` address)

This is no longer true when we use alloc.

openjdk/mmtkHeap.cpp
javadamiri: There are two basic issues here:
1. I don't think this method was ever called before, because the method was not virtual in the parent class. I guess the reason why this hasn't caused any problems so far is that the implementation of this method is equivalent to its implementation in the parent class.
2. Even if we adding a branch here, there will still be performance penalty, because the branch will look like: `if !matches!(SINGLETON.plan.options().plan, PlanSelector::MarkSweep)`, which is not optimisable at compile time. If we still could use the `if !cfg(feature = "marksweep")` check, there would be no performance overhead.

PR #49, opened 2021-02-12
MallocMS

paigereeves:
This PR includes changes to the OpenJDK binding to support MallocMS. Related PR to mmtk-core is [#235](https://github.com/mmtk/mmtk-core/pull/235).

### DIFF
--- a/mmtk/src/api.rs
+++ b/mmtk/src/api.rs
@@ -100,7 +100,7 @@ pub extern "C" fn alloc_slow_bump_monotone_immortal(
 
 #[no_mangle]
 pub extern "C" fn is_in_reserved_malloc(obj: ObjectReference) -> bool {
-    if !cfg!(feature = "marksweep") {
+    if !matches!(SINGLETON.plan.options().plan, PlanSelector::MarkSweep) {
         false
     } else if is_alloced_by_malloc(obj) {
         true
--- a/mmtk/src/api.rs
+++ b/mmtk/src/api.rs
@@ -16,6 +16,7 @@ use crate::OpenJDK;
 use crate::OpenJDK_Upcalls;
 use crate::SINGLETON;
 use crate::UPCALLS;
+use mmtk::util::alloc::is_alloced_by_malloc;
 
 // Supported barriers:
 static NO_BARRIER: SyncLazy<CString> = SyncLazy::new(|| CString::new("NoBarrier").unwrap());
@@ -83,7 +84,7 @@ pub extern "C" fn get_allocator_mapping(allocator: AllocationSemantics) -> Alloc
 // Allocation slow path
 
 use mmtk::util::alloc::Allocator as IAllocator;
-use mmtk::util::alloc::{BumpAllocator, LargeObjectAllocator};
+use mmtk::util::alloc::{BumpAllocator, LargeObjectAllocator, MallocAllocator};
 use mmtk::util::heap::MonotonePageResource;
 
 #[no_mangle]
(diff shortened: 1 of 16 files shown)


mmtk/src/api.rs
```diff
@@ -1,32 +1,37 @@
...
-use crate::UPCALLS;
+use std::ffi::{CStr, CString};
+use std::lazy::SyncLazy;
 
 // Supported barriers:
 static NO_BARRIER: SyncLazy<CString> = SyncLazy::new(|| CString::new("NoBarrier").unwrap());
 static OBJECT_BARRIER: SyncLazy<CString> = SyncLazy::new(|| CString::new("ObjectBarrier").unwrap());
 
+extern {
+    static mut mmtk_enable_allocation_fastpath: bool;
```
qinsoon, line 26: This is unnecessary. We can keep the flag `MMTK_ENABLE_ALLOCATION_FASTPATH`. For malloc allocator, we dont need to implement a fastpath. We just need to insert a slowpath call.

https://github.com/mmtk/mmtk-openjdk/blob/75296ac4300d405e38825c41e105404d71dd8a38/openjdk/mmtkBarrierSetAssembler_x86.cpp#L45
If it is malloc allocator, just jump to slow case.

https://github.com/mmtk/mmtk-openjdk/blob/75296ac4300d405e38825c41e105404d71dd8a38/openjdk/mmtkBarrierSetC2.cpp#L125
Move this line slightly earlier. If the allocator is malloc allocator, we set `always_slow` to `true` (like this: https://github.com/mmtk/mmtk-openjdk/blob/75296ac4300d405e38825c41e105404d71dd8a38/openjdk/mmtkBarrierSetC2.cpp#L80)

mmtk/src/api.rs
```diff
@@ -37,6 +42,9 @@ pub extern "C" fn release_buffer(ptr: *mut Address, length: usize, capacity: usi
 
 #[no_mangle]
 pub extern "C" fn openjdk_gc_init(calls: *const OpenJDK_Upcalls, heap_size: usize) {
+    if cfg!(feature = "marksweep") {
```
qinsoon, line 45: As we now allow run-time plan, it is not necessary to pass in a plan feature for OpenJDK. This would be incorrect if we do not build with `marksweep`.

mmtk/src/api.rs
```diff
@@ -97,6 +105,17 @@ pub extern "C" fn alloc_slow_bump_monotone_immortal(
     unsafe { &mut *(allocator as *mut BumpAllocator<OpenJDK>) }.alloc_slow(size, align, offset)
 }
 
+#[no_mangle]
+pub extern "C" fn is_in_reserved_malloc(obj: ObjectReference) -> bool {
+    if !cfg!(feature = "marksweep") {
```
qinsoon, line 110: We probably should add an API function to check if an object reference is in mmtk heap. This is internal implementation - we should not expose this and let a binding to deal with it.

qinsoon (changes requested):
Can you also add marksweep tests to `ci-test.sh`? We should check what benchmarks can run with marksweep and include them as many as we can.

mmtk/src/api.rs
wenyuzhao: It was me suggesting Paige to turn this constant into a global variable, and enable/disable it at vm boot time, depending on the selected plan. I think this is a more generic solution compare to adding a branch in assembler and c2 IR, especially when we have multiple plans that can't support allocation fast path in the future.

qinsoon: Whether we have fastpath or not depends on the allocator type, not the plan. For example, for MS, we may allocate immortal objects via bump pointer allocator. In this case, we can still use bump pointer fastpath.  (I understand that for the current MS PR in mmtk-core, all allocation is done through malloc, which seems incorrect)

I didn't suggest working around the cases where fastpath is not *yet* supported. What I suggested is to properly implement for the malloc allocator, which basically has no fastpath - all we need is a call to the slowpath. Please correct me if I am wrong.

qinsoon:
I merged this from Paige's fork to the `marksweep` branch in our repo.

PR #50, opened 2021-02-18
Side-metadata supported object remembering barrier

wenyuzhao:
This PR adds object barrier fast-path backed by side-metadata.

https://github.com/mmtk/mmtk-core/pull/184


openjdk/barriers/mmtkObjectBarrier.hpp
```diff
@@ -11,6 +11,13 @@
 #include "../mmtkBarrierSetC1.hpp"
 #include "../mmtkBarrierSetC2.hpp"
 
+#define MMTK_ENABLE_OBJECT_BARRIER_FASTPATH true
+
+#define SIDE_METADATA_BASE_ADDRESS 0x0000060000000000L
```
qinsoon, line 16: This is hardcoded based on https://github.com/mmtk/mmtk-core/blob/a1cda07a572385d25084ad595ae302da888d8bf3/src/util/side_metadata/constants.rs#L7. Any chance we can use the constant from mmtk-core?

@javadamiri Does this need to be changed with your new side metadata implementation?

qinsoon (commented):
We use `slowdebug` in our CI. Is the fastpath tested in our CI?

openjdk/barriers/mmtkObjectBarrier.hpp
wenyuzhao: > We use `slowdebug` in our CI. Is the fastpath tested in our CI?

Yes fast-path is always included. But I suggest to use fastdebug in our CI to speed up the test time. fastdebug builds enabled all assertions as well.

wenyuzhao: Actually all fast-path are hard-coded right now. The fast-path IR only performs correctly under the current MetadataSpec settings. I'm planning to pass the `MetadataSpec` instance to vm-bindings in the future, so VM bindings can generate correct fast-path (as well as these constants) without hardcoding anything.

qinsoon: > > We use `slowdebug` in our CI. Is the fastpath tested in our CI?
> 
> Yes fast-path is always included. But I suggest to use fastdebug in our CI to speed up the test time. fastdebug builds enable all assertions as well.

I changed the build to fastdebug in https://github.com/mmtk/mmtk-openjdk/pull/52, as we will start testing 3 plans since that PR - it is worth spending a bit more time in building and saving time for running tests.

qinsoon: > Actually all fast-path are hard-coded right now. The fast-path IR only performs correctly under the current MetadataSpec settings. I'm planning to pass the `MetadataSpec` instance to vm-bindings in the future, so VM bindings can generate correct fast-path (as well as these constants) without hardcoding anything.

OK. I think we should have @javadamiri to look at this PR as well, as he will change  the side metadata implementation in mmtk-core. It would be better if he knows what kind of assumptions/hardcode we have for the side metadata implementation in this binding PR.

javadamiri: The base metadata address changes to `HEAP_END` in the new implementation, we will need to change this constant.
It also doesn't use chunks (e.g. `CHUNK_MASK`).
Overall, I don't think much change is required on the binding side, except for the fastpath IR which will become more similar to the Wenyu's own implementation.

qinsoon (approved):

openjdk/barriers/mmtkObjectBarrier.hpp
qinsoon: Sounds good. I will start merging the PR in mmtk-core and this PR.

PR #55, opened 2021-03-17
update fastpath to the new side metadata implementation

javadamiri:
This PR is related to mmtk-core issue #261 on the performance regression in the chunk-based implementation of the side metadata.

### DIFF
--- a/mmtk/Cargo.toml
+++ b/mmtk/Cargo.toml
@@ -29,6 +29,7 @@ mmtk = { git = "ssh://git@github.com/mmtk/mmtk-core.git", rev = "0985c60a1d8e430
 default = []
 nogc_lock_free = ["mmtk/nogc_lock_free"]
 nogc_no_zeroing = ["mmtk/nogc_no_zeroing"]
+side_gc_header = ["mmtk/side_gc_header"]
 
 # We can select plan at runtime. So no need to use any of these features.
 # However, if any of these is provided during build-time, we will ignore any runtime flag and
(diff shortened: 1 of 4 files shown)


mmtk/src/api.rs
```diff
@@ -46,7 +46,9 @@ pub extern "C" fn openjdk_gc_init(calls: *const OpenJDK_Upcalls, heap_size: usiz
         unsafe { &mut *(&*SINGLETON as *const MMTK<OpenJDK> as *mut MMTK<OpenJDK>) };
     memory_manager::gc_init(singleton_mut, heap_size);
     if let PlanSelector::GenCopy = SINGLETON.options.plan {
-        assert!(VMObjectModel::HAS_GC_BYTE, "Side GC-byte conflicts with logging bit metadata.");
+        if cfg!(feature = "side_gc_header") {
```
qinsoon, line 49: This seems suggesting `side_gc_header` won't work for GenCopy. Is this only for the OpenJDK binding? Theoratically if we have GenCopy in other binding, does it work?

mmtk/Cargo.toml
```diff
@@ -21,14 +21,15 @@ lazy_static = "1.1"
...
-mmtk = { git = "ssh://git@github.com/mmtk/mmtk-core.git", rev = "b85f621f62c7390029c09e2e136d95e54feea8aa" }
+mmtk = { git = "ssh://git@github.com/mmtk/mmtk-core.git", rev = "0985c60a1d8e4305449731ffc37244f17cc05a44" }
 # Uncomment the following to build locally
 # mmtk = { path = "../repos/mmtk-core" }
 
 [features]
 default = []
 nogc_lock_free = ["mmtk/nogc_lock_free"]
 nogc_no_zeroing = ["mmtk/nogc_no_zeroing"]
+side_gc_header = ["mmtk/side_gc_header"]
```
qinsoon, line 32: Have we tested that `side_gc_header` works for other plans? Otherwise, the only reason we need this feature is to throw a panic to tell people dont use it for GenCopy while we don't know if it works for other plans.

qinsoon (approved):

mmtk/Cargo.toml
javadamiri: I removed that assertion, so we do not use the feature anymore.
I will remove the feature for now

PR #73, opened 2021-05-28
Remove MMTK_PLAN from CompileThirdPartyHeap

caizixian:
Since we have runtime plan selection support now, this option doesn't make sense anymore.


qinsoon:
The features should also be removed from `Cargo.toml` and the code. Besides, we need to 1) merge Wenyu's changes to the running script to support env vars per build and 2) modify the configs in `ci-perf-kit` to select plans by env vars.

caizixian:
> The features should also be removed from `Cargo.toml` and the code. Besides, we need to 1) merge Wenyu's changes to the running script to support env vars per build and 2) modify the configs in `ci-perf-kit` to select plans by env vars.

Agreed.

PR #81, opened 2021-06-23
Alloc to LOS for large objects

qinsoon:
This PR adds size check so we can allocate large objects directly to large object space.
* Adds a size check in the slowpath
* Adds a size check in the fastpath to jump to the slowpath

### DIFF
--- a/openjdk/mmtkBarrierSetAssembler_x86.cpp
+++ b/openjdk/mmtkBarrierSetAssembler_x86.cpp
@@ -36,29 +36,28 @@
 void MMTkBarrierSetAssembler::eden_allocate(MacroAssembler* masm, Register thread, Register obj, Register var_size_in_bytes, int con_size_in_bytes, Register t1, Label& slow_case) {
   assert(obj == rax, "obj must be in rax, for cmpxchg");
   assert_different_registers(obj, var_size_in_bytes, t1);
-  // printf("eden_allocate\n");
+
   if (!MMTK_ENABLE_ALLOCATION_FASTPATH) {
     __ jmp(slow_case);
   } else {
     // We only use LOS or the default allocator. We need to check size
     // max non-los size
-    int max_non_los_bytes = (int)get_max_non_los_default_alloc_bytes();
+    size_t max_non_los_bytes = get_max_non_los_default_alloc_bytes();
     // For size larger than max non-los size, we always jump to slowpath.
     if (var_size_in_bytes == noreg) {
+      assert(con_size_in_bytes <= 128*1024, "con_size_in_bytes should be smaller than 128K");
       // const size: this check seems never true, which means openjdk may have its own size check.
-      if (con_size_in_bytes > max_non_los_bytes) {
-        printf("consize jump: %d > %d\n", con_size_in_bytes, max_non_los_bytes);
-        assert(false, "we are actually jumping to slow due to mmtk size check");
+      if ((size_t)con_size_in_bytes > max_non_los_bytes) {
+        // printf("consize jump: %d > %ld\n", con_size_in_bytes, max_non_los_bytes);
+        // assert(false, "we are actually jumping to slow due to mmtk size check");
         __ jmp(slow_case);
         return;
       }
     } else {
       assert(var_size_in_bytes->is_valid(), "var_size_in_bytes is not noreg, and is not valid");
-      // // var size
-      // printf("varsize\n");
-      // __ cmpptr(var_size_in_bytes, max_non_los_bytes);
-      // __ jcc(Assembler::aboveEqual, slow_case);
-      // return;
+      // var size
+      __ cmpptr(var_size_in_bytes, max_non_los_bytes);
+      __ jcc(Assembler::aboveEqual, slow_case);
     }
 
     // fastpath, we only use default allocator
--- a/openjdk/mmtkBarrierSetAssembler_x86.cpp
+++ b/openjdk/mmtkBarrierSetAssembler_x86.cpp
@@ -40,9 +40,27 @@ void MMTkBarrierSetAssembler::eden_allocate(MacroAssembler* masm, Register threa
   if (!MMTK_ENABLE_ALLOCATION_FASTPATH) {
     __ jmp(slow_case);
   } else {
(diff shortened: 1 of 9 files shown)


openjdk/mmtkMutator.cpp
```diff
@@ -1,12 +1,19 @@
 
 #include "mmtk.h"
 #include "mmtkMutator.hpp"
+#include <atomic>
```
javadamiri, line 4: Where is this used?

qinsoon: Ah. I was doing some instrumentation with atomic integers. I will the include. Sorry.

openjdk/mmtkBarrierSetAssembler_x86.cpp
```diff
@@ -36,13 +36,32 @@
...
   } else {
-    // We always use the default allocator.
-    // But we need to figure out which allocator we are using by querying MMTk.
-    AllocatorSelector selector = get_allocator_mapping(AllocatorDefault);
+    // MMTk size check. If the alloc size is larger than the allowed max size for non los,
+    // we jump to slow path and allodate with LOS in slowpath.
+    // Note that OpenJDK has a slow path check. Search for layout_helper_needs_slow_path and FastAllocateSizeLimit.
+    // I tried to set FastAllocateSizeLimit in MMTkHeap::initialize(). But there are still large objects allocated into the
+    // default space.
+    size_t max_non_los_bytes = get_max_non_los_default_alloc_bytes();
```
javadamiri, line 48: If we had a constant, instead of this function call, would that affect performance?

qinsoon: I measured it, and there was no measurable difference. I think among all these compiler code, this is just one call (calling to Rust has no extra overhead). So it doesn't matter much. And it is not actually a constant though, it depends on the selected plan.

wenyuzhao: This code will only be executed by the assembler when it compiles some Java code. So it will become a compile-time inlined constant in the generated machine code.

wenyuzhao: This will be fired at each Java code compilation. Is it possible to cache `max_non_los_bytes` into a global variable? so we only need to call `get_max_non_los_default_alloc_bytes` once at the beginning. We are not switching plans at runtime, so `get_max_non_los_default_alloc_bytes`  should always return the same value.

wenyuzhao: But I feel that we can't really measure the difference. So having the call inside assembler is also fine.

openjdk/mmtkBarrierSetC2.cpp
```diff
@@ -67,12 +67,61 @@ void MMTkBarrierSetC2::expand_allocate(
...
+
+  // We will do MMTk size check.
+  // The implementation tries not to change the current control flow, and instead evaluate allocation size
+  // with max non-los bytes and combine the result with initial_slow_test. If the size check fails, initial_slow_test
+  // should be evaluated to true, and we jump to the slowpath.
+
+  // The max non-los bytes from MMTk
+  int max_non_los_bytes = (int)get_max_non_los_default_alloc_bytes();
+  // Check if allocation size is constant
+  int const_size = x->_igvn.find_int_con(size_in_bytes, -1);
```
wenyuzhao, line 85: Looks like `size_in_bytes ` has long type (line 99 below). So use `find_int_con` instead of `find_long_con` here may make this optimization never triggered?

openjdk/mmtkBarrierSetC2.cpp
```diff
@@ -67,12 +67,61 @@ void MMTkBarrierSetC2::expand_allocate(
...
+    x->transform_later(mmtk_size_cmp);
+    // Size Check: if size_in_bytes >= max_non_los_bytes
+    Node *mmtk_size_bool = new BoolNode(mmtk_size_cmp, BoolTest::ge);
+    x->transform_later(mmtk_size_bool);
+
+    if (initial_slow_test == NULL) {
+      // If there is no previous initial_slow_test, we use the size check as initial_slow_test
+      initial_slow_test = BoolNode::make_predicate(mmtk_size_bool, &x->_igvn);
+    } else {
+      // If there is existing initial_slow_test, we combine the result by 'or' them together.
```
wenyuzhao, line 109: Is it the openjdk internal slow test you mentioned before? Can we ignore it and do the slow test completely ourselves?

openjdk/mmtkBarrierSetC2.cpp
qinsoon: Good point. I will fix this.

openjdk/mmtkBarrierSetC2.cpp
qinsoon: Yes, that's the one I mentioned before. I don't think we can ignore their test. We jump to the slowpath for larger object size, but they jump to the slowpath for various reasons. 
For example the code [here]( https://github.com/mmtk/openjdk/blob/e1e420b3c4fd627218e45ec977972561c559acec/src/hotspot/share/opto/graphKit.cpp#L3582), they check if a bit is set for the type. But that function takes an argument of `extra_slow_test`, if the bit is not set, they will use the `extra_slow_test`. One example of the `extra_slow_test` is [this](https://github.com/mmtk/openjdk/blob/e1e420b3c4fd627218e45ec977972561c559acec/src/hotspot/share/opto/library_call.cpp#L2864), in which they check whether the class is initialized.

I don't feel confident that we can ignore their test. Also this block is only for variable allocation size which should be a rare case anyway.

openjdk/mmtkBarrierSetAssembler_x86.cpp
qinsoon: I created a static variable in `MMTkMutatorContext` to cache the value, the value is initialized right after we call `gc_init()`. When we read the value, there are assertions to make sure the value is initialized. But I don't think it matters for performance.

wenyuzhao (approved):
LGTM

PR #83, opened 2021-06-25
Test PageProtect GC with higher linux mapping limit

qinsoon:


qinsoon:
This draft PR was just testing if we can increase mappling limit. The changes are included in https://github.com/mmtk/mmtk-openjdk/pull/75, and got merged.

PR #86, opened 2021-07-05
Remove hardcoded base address for VM global side metadata

k-sareen:
Related to https://github.com/mmtk/mmtk-core/pull/361. The VM base address was a hardcoded macro before. Now it directly uses the MMTk core defined constant and so we do not have to mess around with change the constant whenever we update it.


qinsoon (approved):
LGTM

PR #89, opened 2021-07-14
Add feature to switch location of mark bit

k-sareen:
This adds a feature `mark_bit_in_header` that switches the location of the mark bit. This feature requires https://github.com/mmtk/mmtk-core/pull/361 as it uses the newly defined `offset()` function.


k-sareen:
@qinsoon I'm not super familiar with the OpenJDK Makfile etc. How did we want to add the compile time flag to the CI? As it stands, the current `CompileThirdPartyHeap.gmk` uses environment variables to set features.

qinsoon:
> @qinsoon I'm not super familiar with the OpenJDK Makfile etc. How did we want to add the compile time flag to the CI? As it stands, the current `CompileThirdPartyHeap.gmk` uses environment variables to set features.

You can use another env vars as well in CompileThirdPartyHeap.gmk.

k-sareen:
Bit of a hack to make it work nicely. I've also updated the README to reflect that you can choose the location of the mark-bit and the default location. Given that this depends on functions in https://github.com/mmtk/mmtk-core/pull/361, we'll have to merge that one first.

qinsoon:
I think you need to update the mmtk-core dependency to point to your mmtk-core PR. @k-sareen

k-sareen:
I've had to point it to my fork -- we'll have to change it before merging this

qinsoon:
> I've had to point it to my fork -- we'll have to change it before merging this

Yeah. Its fine. We will fix it before merging anyway. The dependency is pointing to one commit in the PR, and the commit will disappear after squashing.

qinsoon (approved):
Looks good. Thanks for the PR.

PR #94, opened 2021-07-22
Allow longer timeout in the CI job. Use larger mapping limit.

qinsoon:


qinsoon:
Closing this, as the munprotect fail was not caused by mapping limit.

PR #102, opened 2021-08-17
Fix monitor use in finalizer thread.

wks:
-   Use a boolean variable to indicate if finalization is scheduled,
    instead of using the monitor alone.

-   Use ThreadBlockInVM instead of checking safepoint at monitor
    lock/wait.


wenyuzhao (approved):
LGTM

PR #105, opened 2021-09-02
Style fixes and minor fixes for C++ code in the OpenJDK binding.

wks:
This PR aims to fix some obvious and easy-to-fix style problems.

TODO list:

-   [X] Include guards (`#ifndef` / `#define`): Make the macros longer and more uniform, using the file-system path as a guideline.
-   [X] Indentation: HotSpot uses 2-space indentation.
-   [x] Indentation: Indent parameter/argument list to the left bracket.
-   [X] Pre-compiled header: `precompiled.hpp` should be the first to `#include` in every `.cpp` files, but never included in any `.hpp` files.
-   [X] Standard libraries: HotSpot avoids use of standard libraries.
-   [X] Reorder `#include`s: First `precompiled.hpp` (`.cpp` only), then unconditional headers alphabetically, and finally conditional headers.
-   [X] Author/date info in code comments: Use `git blame`, instead.

Do these in a different PR:

-   These require some non-trivial refactoring around the `third_party_heap` namespace and two `ThirdPartyHeap.hpp` files:
    -   Namespaces: HotSpot does not use C++ namespaces.
    -   Unique basename: HotSpot's build system requires all source files to have globally unique basename.
-   Cannot decide the purpose of the code commented out.
    -   Dead code: Remove commented code.

HotSpot style guide: https://github.com/openjdk/jdk/blob/master/doc/hotspot-style.md


openjdk/mmtkRootsClosure.hpp
```diff
@@ -134,5 +134,4 @@ class MMTkScanObjectClosure : public BasicOopIterateClosure {
 //   virtual void do_cld(ClassLoaderData* cld) {
 
 //     printf("CLD: %p", p);
-//   }
-// };
\ No newline at end of file
+// };//   }
```
qinsoon, line 137: I don't know where the second `// }` is from. Might be a mistake from regex replacement.

qinsoon (approved):
Looks good

openjdk/mmtkRootsClosure.hpp
wks: In that commit, I intended to normalize all `#ifndef`, `#define` and `#endif` directives.  I edited all `.hpp` files with `sed`.  However, it looks like this file `mmtkRootsClosure.hpp` uses CRLF line ending instead of LF, and may have confused `sed`.

I will normalize line endings, too.

By the way, I find that `mmtkRootsClosure.hpp` does not have any `#ifndef` guards. I'll fix that, too.

wks: I remember what happened.

1.  That file does not end with any newline characters.
     -   NOTE: Older C/C++ specs require all source files to end with a newline in case some C/C++ preprocessor `#include` a file that does not end with `\n` or `\r\n` will corrupt the next line following the `#include`, but newer C/C++ specs no longer require it.
     -   VSCode, by default, does not insert newline char at the end of a file. But VSCode can be set to do so in the setting.
     -   Vim, Emacs and all traditional text editors automatically insert newline at the end of a file.
2.  My algorithm to match the last `#endif` is to
    1. Reverse all lines with `tac`.
    2. Then use `sed` to match the now "first" `#endif`
    3. Then use `tac` to reverse all lines again.

Because the file did not end with a newline character, the `tac` program got confused, and joined the last two lines in the reverse order.

I think it is worth enforcing UNIX-style line ending (LF) and requiring all files to end with a newline character.

PR #108, opened 2021-09-18
Fix stop-the-world race.

wks:
This PR fixes the race bug during stop-the-world.

This PR has two associated PRs:
-   `openjdk`: https://github.com/mmtk/openjdk/pull/9
-   `mmtk-core`: https://github.com/mmtk/mmtk-core/pull/450

In short, the bug exists because the MMTk coordinator calls `SafepointSynchronnize::begin()` directly.  This races with the real VM thread.

Fixing this involves introducing a dedicated "companion thread".  This thread receives stop-the-world requests from the MMTk core. It will let the VM thread execute a VM Operation which does nothing but waiting for another start-the-world request from the MMTk.  In this way, mmtk-core can request stop-the-world, then continue to do GC, then request start-the-world.

Concretely, the changes include:

-   `openjdk`: The `SafepointSynchronize::begin()` method no longer reports which threads have stopped, because this method is used in all VMOperations that needs STW, not just GC.
-   `openjdk`: Added a new VMOperation identifier `VMOp_ThirdPartyHeapOperation`.
-   `mmtk-openjdk`: Introduce a `MMTkVMCompanionThread`.  This is a new non-Java thread. The only thing it does, upon request from MMTk core, is to create a `VM_MMTkSTWOperation` and call `VMThread::execute` to execute it.
    The VM Thread then stops the world, executes the `VM_MMTkSTWOperation`. It finishes when the mmtk-core requests start-the-world.  Then the VM Thread will start the world and unblock the companion thread
    The reason why we introduce a companion thread is that `VMThread::execute` will block the caller until the VM thread finished executing the VM operation.  There is an alternative. **We could use "asynchronous" (non-blocking) VM operations, but this feature is removed in a later OpenJDK version**.
-   `mmtk-core`: Because now the only thread that calls `SafepointSynchronize::begin()` and `SafepointSynchronize::end()` is the real VM thread, we no longer need to let the coordinator request stop/start-the-world.

There could be some performance impact because of the companion thread we introduced, so performance tests are needed.  Some preliminary tests show no obvious performance regression.

# Merging strategy

All three repositories should be updated at the same time.

If this is impossible, we should update `mmtk-core` and `openjdk` first. Then apply code changes to `mmtk-openjdk` and, at the same time, update its dependencies so that it depends on the updated `mmtk-core` and `openjdk`.

### DIFF
--- a/openjdk/mmtkVMCompanionThread.cpp
+++ b/openjdk/mmtkVMCompanionThread.cpp
@@ -1,0 +1,138 @@
+/*
+ * Copyright (c) 1998, 2017, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
+
+#include "precompiled.hpp"
+#include "mmtk.h"
+#include "mmtkVMCompanionThread.hpp"
+#include "runtime/mutex.hpp"
+
+MMTkVMCompanionThread::MMTkVMCompanionThread():
+    NamedThread(),
+    _desired_state(_threads_resumed),
+    _reached_state(_threads_resumed) {
+  set_name("MMTK VM Companion Thread");
+  _lock = new Monitor(Monitor::nonleaf,
+                      "MMTkVMCompanionThread::_lock",
+                      true,
+                      Monitor::_safepoint_check_never);
+}
+
+MMTkVMCompanionThread::~MMTkVMCompanionThread() {
+  guarantee(false, "MMTkVMCompanionThread deletion must fix the race with VM termination");
+}
+
+void MMTkVMCompanionThread::run() {
+  this->initialize_named_thread();
+
+  for (;;) {
+    // Wait for suspend request
+    log_trace(gc)("MMTkVMCompanionThread: Waiting for suspend request...");
+    {
+      MutexLockerEx locker(_lock, Mutex::_no_safepoint_check_flag);
+      assert(_reached_state == _threads_resumed, "Threads should be running at this moment.");
+      while (_desired_state != _threads_suspended) {
+        _lock->wait(true);
+      }
+      assert(_reached_state == _threads_resumed, "Threads should still be running at this moment.");
+    }
+
+    // Let the VM thread stop the world.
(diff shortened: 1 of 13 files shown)


openjdk/mmtkVMCompanionThread.cpp
```diff
@@ -0,0 +1,118 @@
...
+  _lock->notify_all();
+
+  if (wait_until_reached) {
+    while (_reached_state != desired_state) {
+      _lock->wait(true);
+    }
+  }
+}
+
+void MMTkVMCompanionThread::wait_for_reached(stw_state desired_state) {
```
qinsoon, line 96: This seems not used anywhere.

openjdk/mmtkVMCompanionThread.cpp
```diff
@@ -0,0 +1,118 @@
...
+      while (_desired_state != _threads_suspended) {
+        _lock->wait(true);
+      }
+      assert(_reached_state == _threads_resumed, "Threads should still be running at this moment.");
+    }
+
+    // Let the VM thread stop the world.
+    log_trace(gc)("MMTkVMCompanionThread: Letting VMThread execute VM op...");
+    VM_MMTkSTWOperation op(this);
+    VMThread::execute(&op);
```
qinsoon, line 64: Is `execute()` blocking? We probably want to have some comments here.
```suggestion
    // execute() is blocking. The companion thread will be blocked here and the VM thread will be blocked in reach_suspended_and_wait_for_resume(), until a request() for resume comes in.
    VMThread::execute(&op);
```

openjdk/mmtkVMCompanionThread.hpp
```diff
@@ -0,0 +1,58 @@
...
+public:
+  enum stw_state {
+    _threads_suspended,
+    _threads_resumed,
+  };
+private:
+  Monitor* _lock;
+  stw_state _desired_state;
+  stw_state _reached_state;
+  size_t _resumption_count;
```
qinsoon, line 44: This count is maintained but not used.

qinsoon (approved):
LGTM

openjdk/mmtkCollectorThread.hpp
```diff
@@ -39,10 +39,6 @@ class MMTkCollectorThread: public NamedThread {
     guarantee(false, "MMTkCollectorThread deletion must fix the race with VM termination");
   }
 
-  // Tester
-  bool is_VM_thread() const                      { return true; }
```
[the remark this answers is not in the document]
wks: That's right.  There is only one VM thread, i.e. the `VMThread` class defined in `src/hotspot/share/runtime/vmThread.hpp`.

openjdk/mmtkVMCompanionThread.cpp
wks: For JDK 11, not always.  It depends on the `evaluation_mode()` and `evaluation_concurrently()`.  There is only one non-blocking VMOperation in JDK 11.  The `VM_MMTkSTWOperation` operation I defined is blocking.

For JDK 17, all VM operations are blocking.

Anyway I think it is helpful to add a comment.

openjdk/mmtkVMCompanionThread.cpp
wks: My intention was to allow two different ways to use the `request` method:

1.  Call `request(some_state, true)`. The thread will block until reaching `some_state`.
2.  Call `request(some_state, false)`, do something else, and then call `wait_for_reached(some_state)`.  This allows some things to be done while the VM is stopping threads.

Currently in `mmtkUpcalls.cpp`, `mmtk_stop_all_mutators` and `mmtk_resume_mutators` only uses the first way, because this is the interface mmtk-core requires the VM to implement.  But I think it may be helpful in the future to allow this usage.

openjdk/mmtkVMCompanionThread.hpp
wks: I once considered moving the `start_the_world_count` into the companion thread as `_resumption_count`, but it didn't work as expected.  I will remove `_resumption_count`.

wks:
I made some updates. Added comments, removed stale code, and added "MMTk" before the names of the context and the collector threads.

PR #113, opened 2021-09-27
Fix Issue 112

### LINKED ISSUE #112: Assertion fail when alloc bit is enabled
binding: v0.7.0 145acad3d617ca29d1fa54f64f7e2c83ebd65452

Build:
```
export GLOBAL_ALLOC_BIT=1 && export DEBUG_LEVEL=slowdebug && export MMTK_PLAN= && export RUSTUP_TOOLCHAIN=nightly-2021-09-17 && sh configure --disable-warnings-as-errors --with-debug-level=$DEBUG_LEVEL && make CONF=linux-x86_64-normal-server-$DEBUG_LEVEL THIRD_PARTY_HEAP=$PWD/../../openjdk
```

Run:
```
$ MMTK_PLAN=NoGC build/linux-x86_64-normal-server-slowdebug/jdk/bin/java -XX:+UseThirdPartyHeap[2021-09-24T07:08:57Z INFO  mmtk::memory_manager] Initialized MMTk with NoGC
# To suppress the following error report, specify this argument
# after -XX: or in .hotspotrc:  SuppressErrorAt=/register.hpp:160
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  Internal Error (/home/yilin/Code/mmtk-openjdk/repos/openjdk/src/hotspot/share/asm/register.hpp:160), pid=11466, tid=11501
#  assert(a != b && a != c && a != d && a != e && b != c && b != d && b != e && c != d && c != e && d != e) failed: registers must be different: a=0x0000000000000000, b=0x000000000000000a, c=0x000000000000000b, d=0x0000000000000001, e=0x0000000000000001
#
# JRE version: OpenJDK Runtime Environment (11.0.11) (slowdebug build 11.0.11-internal+0-adhoc.yilin.openjdk)
# Java VM: OpenJDK 64-Bit Server VM (slowdebug 11.0.11-internal+0-adhoc.yilin.openjdk, mixed mode, tiered, third-party gc, linux-amd64)
# Core dump will be written. Default location: Core dumps may be processed with "/usr/share/apport/apport %p %s %c %d %P %E" (or dumping to /home/yilin/Code/mmtk-openjdk/repos/openjdk/core.11466)
#
# An error report file with more information is saved as:
# /home/yilin/Code/mmtk-openjdk/repos/openjdk/hs_err_pid11466.log
#
# Compiler replay data is saved as:
# /home/yilin/Code/mmtk-openjdk/repos/openjdk/replay_pid11466.log
#
# If you would like to submit a bug report, please visit:
#   https://bugreport.java.com/bugreport/crash.jsp
#
Current thread is 11501
Dumping core ...
Aborted (core dumped)
```

The assertion is in our fastpath compilation: https://github.com/mmtk/mmtk-openjdk/blob/145acad3d617ca29d1fa54f64f7e2c83ebd65452/openjdk/mmtkBarrierSetAssembler_x86.cpp#L118

The issue does not show in our CI, as the CI scripts add `-XX:-TieredCompilation` so we are not executing the fastpath generation.

YiluoWei:
fix #112

change the command line arguments in ci-tests, so that they run in C1-C2 tiered compilation mode (by default).

### DIFF
--- a/openjdk/mmtkBarrierSetAssembler_x86.cpp
+++ b/openjdk/mmtkBarrierSetAssembler_x86.cpp
@@ -112,12 +112,9 @@ void MMTkBarrierSetAssembler::eden_allocate(MacroAssembler* masm, Register threa
     __ movptr(cursor, end);
 
 #ifdef MMTK_ENABLE_GLOBAL_ALLOC_BIT
-    Register tmp3 = rscratch2;
+    Register tmp3 = rdi;
     Register tmp2 = rscratch1;
     assert_different_registers(obj, tmp2, tmp3, rcx);
-
-    __ push(tmp2);
-    __ push(tmp3);
 
     // tmp2 = load-byte (SIDE_METADATA_BASE_ADDRESS + (obj >> 6));
     __ movptr(tmp3, obj);
@@ -141,8 +138,6 @@ void MMTkBarrierSetAssembler::eden_allocate(MacroAssembler* masm, Register threa
     __ movptr(rcx, ALLOC_BIT_BASE_ADDRESS);
     __ movb(Address(rcx, tmp3), tmp2);  
 
-    __ pop(tmp3);
-    __ pop(tmp2);
 #endif
 
     // BarrierSetAssembler::incr_allocated_bytes
--- a/openjdk/mmtkBarrierSetAssembler_x86.cpp
+++ b/openjdk/mmtkBarrierSetAssembler_x86.cpp
@@ -114,8 +114,10 @@ void MMTkBarrierSetAssembler::eden_allocate(MacroAssembler* masm, Register threa
 #ifdef MMTK_ENABLE_GLOBAL_ALLOC_BIT
     Register tmp3 = rscratch2;
     Register tmp2 = rscratch1;
-    Register tmp4 = t1;
-    assert_different_registers(obj, tmp2, tmp3, tmp4, rcx);
+    assert_different_registers(obj, tmp2, tmp3, rcx);
+
+    __ push(tmp2);
+    __ push(tmp3);
 
     // tmp2 = load-byte (SIDE_METADATA_BASE_ADDRESS + (obj >> 6));
     __ movptr(tmp3, obj);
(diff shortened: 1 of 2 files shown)


openjdk/mmtkBarrierSetAssembler_x86.cpp
```diff
@@ -114,33 +114,35 @@ void MMTkBarrierSetAssembler::eden_allocate(MacroAssembler* masm, Register threa
 #ifdef MMTK_ENABLE_GLOBAL_ALLOC_BIT
     Register tmp3 = rscratch2;
     Register tmp2 = rscratch1;
-    Register tmp4 = t1;
-    assert_different_registers(obj, tmp2, tmp3, tmp4, rcx);
+    assert_different_registers(obj, tmp2, tmp3, rcx);
+
+    __ push(tmp2);
```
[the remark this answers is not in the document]
YiluoWei: Yes, we can destroy the values in rscratch1/2 if the caller saves their values (the caller should save them if it wants to use them after calling another function). However, in this case I find that I have to save rscratch2, otherwise something will be broken.

Anyway, I now expicitly use rdi here instead of rscratch2.

openjdk/mmtkBarrierSetAssembler_x86.cpp
```diff
@@ -112,35 +112,32 @@ void MMTkBarrierSetAssembler::eden_allocate(MacroAssembler* masm, Register threa
...
-    //   2. tmp4 = 1 << tmp4
-    __ movptr(tmp3, rcx);
-    __ movl(rcx, tmp4);
-    __ movptr(tmp4, 1);
-    __ shlptr(tmp4);
-    __ movptr(rcx, tmp3);
-    // tmp2 = tmp2 | tmp4
-    __ orptr(tmp2, tmp4);
+    //   1. rcx = (obj >> 3) & 7
+    __ movptr(rcx, obj);
```
qinsoon, line 126: This destroys the value in rcx. Do we know that we can safely destroy the value in rcx?

YiluoWei: Yes. rcx is used to pass the argument to a function, and should be saved by the caller if necessary.

qinsoon: It is not related with whether rcx is caller saved or not. But on second thought, it is probably fine, as in this code generation pass, we are still using temporary registers. That suggests the register allocation is done after this pass. We should be fine for using physical registers and the register allocator will deal with it.

qinsoon (approved):
LGTM

PR #114, opened 2021-09-29
No longer restrict STW to coordinator

wks:
After introducing the companion thread, stop-the-world can be triggered
by any GC thread.  This commit makes use of the change in mmtk-core to
elide one redundant context switch.


qinsoon (approved):
LGTM

PR #115, opened 2021-10-01
Fix derived root pointers

wenyuzhao:


openjdk/mmtkUpcalls.cpp
```diff
@@ -49,6 +49,12 @@ static void mmtk_stop_all_mutators(void *tls, void (*create_stack_scan_work)(voi
   MMTkHeap::heap()->companion_thread()->request(MMTkVMCompanionThread::_threads_suspended, true);
   log_debug(gc)("Mutators stopped. Now enumerate threads for scanning...");
 
+  ClassLoaderDataGraph::clear_claimed_marks();
```
wks, line 52: These added lines of code happen after all threads stopped, and before we create any work packets for stack scanning.  Is this necessary?

If we allow more "eager" stack scanning, i.e. create one work packet for each thread when the thread is visited during the `companion_thread()->request()` above, these lines seem to have no place to fit in.

wenyuzhao: Fixed. This code is moved before making the `_threads_suspended` request.

openjdk/mmtkUpcalls.cpp
```diff
@@ -266,6 +279,16 @@ static size_t mmtk_number_of_mutators() {
   return Threads::number_of_threads();
 }
 
+static void mmtk_prepare_for_sanity_roots_scanning() {
+#if COMPILER2_OR_JVMCI
+  DerivedPointerTable::update_pointers();
```
qinsoon, line 284: Normally it is required to do `clear()` before GC, and `update_pointers()` after GC. Why do we need to call these two methods in a 'reverse' order? Are we trying to restore the state of `DerivedPointerTable` to what it is before GC so we can do another stack scanning correctly?

PR #124, opened 2021-11-04
Mark compact

tianleq:


qinsoon:
Can you update the mmtk dependency in `Cargo.toml` to point to your latest commit in the mmtk-core PR? @tianleq

qinsoon:
You will need to change the git repo in the dependency as well. The commit is in your fork. @tianleq

tianleq:
> You will need to change the git repo in the dependency as well. The commit is in your fork. @tianleq

Yeah, I realized that and I am monitoring the result as well

openjdk/mmtkBarrierSetAssembler_x86.cpp
```diff
@@ -111,7 +113,8 @@ void MMTkBarrierSetAssembler::eden_allocate(MacroAssembler* masm, Register threa
     // lab.cursor = end
     __ movptr(cursor, end);
 
-#ifdef MMTK_ENABLE_GLOBAL_ALLOC_BIT
+// #ifdef MMTK_ENABLE_GLOBAL_ALLOC_BIT
```
qinsoon, line 116: With this change, for plans with alloc bit on (e.g. semispace with alloc bit turned on), you are not generating alloc bit code for them. This seems incorrect. But I have no idea how the tests passed. Any idea? @YiluoWei

YiluoWei: I think before testing the global alloca bit, the CI tests already failed.

Probably we need some changes in mmtk-core so that when global alloc bit is enabled, the allocator tag (`TAG_BUMP_POINTER`) in other plans becomes `TAG_BUMP_POINTER_ALLOC_BIT`.

qinsoon: Ah I see. I thought the tests were in the test-assertions. Thanks Yiluo.

openjdk/mmtkBarrierSetC2.cpp
```diff
@@ -229,8 +238,15 @@ void MMTkBarrierSetC2::expand_allocate(PhaseMacroExpand* x,
     Node *old_eden_top = new LoadPNode(ctrl, contended_phi_rawmem, eden_top_adr, TypeRawPtr::BOTTOM, TypeRawPtr::BOTTOM, MemNode::unordered);
 
     x->transform_later(old_eden_top);
+
+    Node *offset = ConLNode::make(extra_header);
+    x->transform_later(offset);
+
+    Node *old_eden_top_plus_offset =  new AddPNode(x->top(), old_eden_top, offset);
+    x->transform_later(old_eden_top_plus_offset);
```
qinsoon, line 246: This code generates the add instruction for all the allocators, which is unnecessary and may have a performance impact. the extra header should only be applied to the mark compact allocator. I am not sure if this is related with the bug though.

tianleq: Yes, but offset is a const node. I presume the impact is negligible. The bug must be related to the offset since for all other allocators with 0 offset, hotspot will not crash.

qinsoon: It will depend on whether the C2 compiler does any optimization after this pass. I once had a `if` based on a const node in this code, and it turned out that the compiler does not optimize that `if`. So even though the offset is a const node, the compiler may actually generates an add instruction.

qinsoon (approved):

PR #132, opened 2022-01-06
Use process_bulk() API and ThirdPartyHeapOptions

qinsoon:


openjdk/mmtkHeap.cpp
```diff
@@ -77,6 +77,12 @@ jint MMTkHeap::initialize() {
   size_t mmtk_heap_size = heap_size;
   /*forcefully*/ //mmtk_heap_size = (1<<31) -1;
 
+  // Set options
+  if (ThirdPartyHeapOptions != NULL) {
```
caizixian, line 81: Once we set the default value to be an empty string, we can get rid of this null check.

caizixian (commented):
See also mmtk/openjdk#10

caizixian (approved):

PR #135, opened 2022-01-29
Fix GC controller thread context

wks:
This PR is associated to https://github.com/mmtk/mmtk-core/pull/533

The problem is stated in https://github.com/mmtk/mmtk-core/issues/532


qinsoon (approved):
LGTM

Issue #78, opened 2021-06-15
Throw `OutOfMemory` instead of panicking


caizixian:
Currently, the OpenJDK binding doesn't override the `out_of_memory` implementation in `Collection`, so when running with extremely small heap size, the VM will panic in Rust instead of throwing a Java exception `OutOfMemory`.

This was done properly in the JikesRVM binding https://github.com/mmtk/mmtk-jikesrvm/blob/8cbdfd6/mmtk/src/collection.rs#L59

k-sareen:
So there is a mismatch between what MMTk expects and what OpenJDK expects to occur on an OOM error.

#### OpenJDK
OpenJDK checks for OOM every (slowpath) allocation [1, 2]. If the object was successfully allocated, it verifies the allocation etc [1]. If the VM is OOM, then a pending exception bit is set for the current thread [2] and the `null` pointer is passed through to the calling Java thread [1]. Here, the Java thread will then check the pending exception bit and then throw the corresponding exception, in this case, it will throw the `OutOfMemoryError`. (One can confirm that it is the Java thread that is throwing the exception instead of the VM as a breakpoint on `print_stack_trace` [3] is not triggered for the `OutOfMemoryError`).

Hence, **OpenJDK allows for allocations to fail (due to OOM) and `null` pointers be passed to Java thread(s)**.

\[1]: https://github.com/mmtk/openjdk/blob/jdk-11.0.11+6-mmtk/src/hotspot/share/gc/shared/memAllocator.cpp#L83
\[2]: https://github.com/mmtk/openjdk/blob/jdk-11.0.11+6-mmtk/src/hotspot/share/gc/shared/memAllocator.cpp#L115-L145
\[3]: https://github.com/mmtk/openjdk/blob/jdk-11.0.11+6-mmtk/src/hotspot/share/classfile/javaClasses.cpp#L2251

#### MMTk
MMTk checks for OOM every (slowpath) allocation [4] as well. However, MMTk handles the OOM case differently. As per the `out_of_memory()` function in `Collection` [5], MMTk expects that an OOM error is thrown immediately. In fact both the binding and the core have many assumptions that MMTk will never return a `null` to the VM for example [6] and [7] to name a few (**Note:** I haven't done a complete audit of the binding and core for this assumption). Naively returning `null` to the VM at an OOM error breaks underlying assumptions of how MMTk works.

Hence, **MMTk's contract with the VM expects the VM to crash the moment an OOM is reported**.

[...]

PR #136, opened 2022-02-01
Properly handle OOM error

k-sareen:
Properly handle the two kinds of OOM error from mmtk-core. Since mmtk-core can return `null` now in the case of an OOM, I've added null checks to some functions. I've already pointed the PR to my mmtk-core PR for testing.

Associated mmtk-core PR: https://github.com/mmtk/mmtk-core/pull/535

Fixes: #78.


k-sareen:
I think maybe we should also add a test for OOM handling to the binding. Let me know if I should add a test.

openjdk/mmtk.h
```diff
@@ -11,6 +11,10 @@ extern "C" {
 
 typedef void* MMTk_Mutator;
 typedef void* MMTk_TraceLocal;
+typedef enum {
```
qinsoon, line 14: Possibly add comments to state that this has the same layout as `mmtk::util::alloc::MmtkAllocationError`.

qinsoon (approved):
LGTM

PR #141, opened 2022-02-22
Fix root scanning

wenyuzhao:
This PR fixes memory leaks caused by incorrect root scanning.

Pointers in `CodeCache`, `StringTable` and `WeakProcessor` are not treated as roots. The processing of these object storage is postponed to the end of tracing. Pointers in these object storages are either removed (if they are dead) or forwarded (if they're moved).

### DIFF
--- a/mmtk/src/api.rs
+++ b/mmtk/src/api.rs
@@ -274,3 +274,22 @@ pub extern "C" fn get_finalized_object() -> ObjectReference {
         None => unsafe { Address::ZERO.to_object_reference() },
     }
 }
+
+/// Test if an object is live at the end of a GC.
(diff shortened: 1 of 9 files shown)


wenyuzhao:
@qinsoon is it possible to trigger a benchmark github action here? like the ones we did to the mmtk-core PRs?

qinsoon:
> @qinsoon is it possible to trigger a benchmark github action here? like the ones we did to the mmtk-core PRs?

Unfortunately we can't. Those are only set up for mmtk-core yet. I can set up performance CI for OpenJDK, as we are doing more development work for OpenJDK and it is probably necessary/worth to set up performance CI.

PR #145, opened 2022-03-15
Weak reference support

qinsoon:
This adds weak reference support, and updates MMTk core to https://github.com/mmtk/mmtk-core/pull/564.

* if weak reference is enabled in MMTk, we will add weak references during our tracing. Otherwise, still treat them as strong references.
* implement new changes in https://github.com/mmtk/mmtk-core/pull/564
* add test with weak reference enabled (by default, it is disabled).


wenyuzhao:
FYI this is the change that fixed the weak processor https://github.com/wenyuzhao/mmtk-openjdk/commit/1bd175fe1fbf053bdd76af4a47ac0c1ea6874ece. This is also a part of https://github.com/mmtk/mmtk-openjdk/pull/141. This change removes the code that treats all the WeakProcessor objects as roots. We should include the change in this PR to deal with weak pointers correctly. Probably this will also yield better performance.

qinsoon:
> FYI this is the change that fixed the weak processor [wenyuzhao@1bd175f](https://github.com/wenyuzhao/mmtk-openjdk/commit/1bd175fe1fbf053bdd76af4a47ac0c1ea6874ece). This is also a part of #141. This change removes the code that treats all the WeakProcessor objects as roots. I'm afraid without this fix, a lot of weak objects will be kept alive. We should include the change in this PR to deal with weak pointers correctly. Probably this will also yield better performance.

@wenyuzhao It seems, to apply the changes in the commit, I need to remove the use of `MMTkRootScanWorkScope`. Otherwise, I will see `must not call oops_do_marking_prologue twice in a row`.

mmtk/src/lib.rs
```diff
@@ -77,6 +77,7 @@ pub struct OpenJDK_Upcalls {
     pub number_of_mutators: extern "C" fn() -> usize,
     pub schedule_finalizer: extern "C" fn(),
     pub prepare_for_roots_re_scanning: extern "C" fn(),
+    pub enqueue_references: extern "C" fn(objects: &[ObjectReference], len: usize),
```
wks, line 80: Slice type `[T]` is not FFI-safe in Rust.  Rust will warn you if you define a function like this:
```rust
extern "C" fn enqueue_references(objects: &[ObjectReference], len: usize) { // WARNING: `extern` fn uses type `[ObjectReference]`, which is not FFI-safe
   ...
}
```

But somehow Rust doesn't give a warning if we define a function pointer (`extern "C" fn(&[T])`) with `extern "C"`. It should give a warning.

From the disassembly, I see that a `&[T]` argument is actually passed via two registers.  On x86_64, `RDI` holds the pointer to the underlying bufer, and `RSI` holds the length of it.  Your code **happens to work as expected** because the actual function 
```C
void mmtk_enqueue_references(void** objects, size_t len)
```
takes a `size_t len` as its second argument, which happens to use the register (`RSI`) for the length of the `&[T]` slice passed from Rust. If you add a third parameter `foo` to it:
```c
void mmtk_enqueue_references(void** objects, size_t len, size_t foo)
```
The third parameter `foo` will receive the actual `len: usize` argument passed from Rust (via register `RDX`).

To properly pass a slice `let s: &[T] = ...` to C, use `s.as_ptr()` to get the pointer, and use `s.len()` to get its length.

mmtk/src/reference_glue.rs
```diff
@@ -1,23 +1,24 @@
...
     }
-    fn process_reference<T: TraceLocal>(
-        _trace: &mut T,
-        _reference: ObjectReference,
-        _tls: VMWorkerThread,
-    ) -> ObjectReference {
-        unimplemented!()
+    fn enqueue_references(references: &[ObjectReference], _tls: VMWorkerThread) {
+        unsafe {
+            ((*UPCALLS).enqueue_references)(references, references.len());
```
wks, line 21: The signature of `OpenJDK_Upcalls::enqueue_references` should be changed.  See the other comment.
```suggestion
            ((*UPCALLS).enqueue_references)(references.as_ptr(), references.len());
```

wks (changes requested):

mmtk/src/lib.rs
qinsoon: Thanks for pointing this out. Fixed.

wks (approved):
LGTM

PR #147, opened 2022-03-30
Moving to stable Rust

qinsoon:
This PR removes use of nightly features.
* `once_cell`: replaced by the `once_cell` crate.
* `vec_into_raw_parts`: replaced by the actual implementation of the method in the binding.
* `box_syntax`: `box X` -> `Box::new(X)`.
* `specialization`: The code used to use `specialization` for type equality in traits `ArrayOopDesc<T>`. 
  * `ArrayOopDesc<T>` is used to describe different Java arrays. In OpenJDK's code base (C++), `arrayOopDesc` has subclasses `objArrayOopDesc` (for arrays of objects) and `typeArrayOopDesc` (for arrays of primitive types). In the binding, `ArrayOopDesc<Oop>` is `objArrayOopDesc`, and `ArrayOopDesc<f64/i64/etc>` is `typeArrayOopDesc`.
  * Now for type equality, an enum for types is used for where type equality is needed.
  * `ArrayOopsDesc` no longer takes any type parameter.
  * `ArrayOopsDesc::data() -> &[T]` takes a Rust type parameter that provides a Rust slice to access the array data.

### DIFF
--- a/mmtk/Cargo.toml
+++ b/mmtk/Cargo.toml
@@ -20,13 +20,14 @@ openjdk_version = "67d5d2b16aacb2ea948552fab2323ebd0abbe924"
 [dependencies]
 libc = "0.2"
 lazy_static = "1.1"
+once_cell = "1.10.0"
 # Be very careful to commit any changes to the following mmtk dependency, as our CI scripts (including mmtk-core CI)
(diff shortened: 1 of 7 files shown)


wks (approved):
LGTM

PR #148, opened 2022-03-30
Replace TransitiveClosure with EdgeVisitor

wks:
In respond to an upstream API change, we now use EdgeVisitor for object
scanning.  This also removes `unreachable!()` implementations for
`process_node()`.

This is associated with the PR in mmtk-core: https://github.com/mmtk/mmtk-core/pull/566


qinsoon (approved):
LGTM

PR #149, opened 2022-03-31
Update allocator struct layouts for stress test

k-sareen:
Adds extra boolean to the allocator structs in order to match the layout in https://github.com/mmtk/mmtk-core/pull/568. I've also updated the local mmtk path in the Cargo.toml file in order to reflect the fact that we don't have a `repos` folder anymore


k-sareen:
Closed in favour of #151

PR #153, opened 2022-04-14
Update OpenJDK to 11.0.15+8-mmtk

k-sareen:


k-sareen:
I added a useful check to the build script that will grab local openjdk version and compare it against the one specified in the cargo metadata in Cargo.toml

qinsoon (approved):
LGTM

PR #155, opened 2022-05-02
Only do the normal build once in binding test

qinsoon:
This PR removes unnecessary builds in the CI script so the script takes less time and is able to finish within 6h limit (Github hosted runner only run a job up to 6h).
* Separates a `ci-build.sh` that builds OpenJDK.
* Use the same build to do normal tests and do weak ref tests.
* Removes some weak reference tests. Now weak reference testing only tests immix and mark compact.


k-sareen:
Maybe we can skirt around the 6 hr requirement by building once and then having different job for each GC (does this mean it can also be parallelized? I am not sure how GitHub Actions work). Though it can be considered over-provisioning of resources.

qinsoon:
> Maybe we can skirt around the 6 hr requirement by building once and then having different job for each GC (does this mean it can also be parallelized? I am not sure how GitHub Actions work). Though it can be considered over-provisioning of resources.

Yeah. That's possible to some extent. We can easily do that in the mmtk-openjdk repo. But for the mmtk-core repo, the current protocol is to invoke `ci-test.sh` in mmtk-openjdk. So we cannot parallelize it.

Anyway, it seems that by removing a few things, we can get the current tests within 6h.

k-sareen:
Ah I see! Proper testing is a hard problem haha. We should look at uphauling our testing infrastructure at some point. I think if parallelization of testing is possible, it would definitely speed up both PR review and developer turnaround time.

k-sareen (approved):
LGTM. Thanks! In the commit message, please mention you're reducing the scope of the weak referencing tests to just Immix and MarkCompact.

Issue #117, opened 2021-10-14
Mutator threads hang


tianleq:
When a mutator tries to allocate memory, it might trigger a gc and the mutator will end up calling ```mmtk_block_for_gc``` in ```mmtkUpcalls.cpp```. Internally, ```mmtk_start_the_world_count``` is used to indicate each round of gc and mutators relies on its value to know which gc to wait for. The issue resides in ```MutexLocker locker(MMTkHeap::heap()->gc_lock());```.  A thread tries to acquire the gc_lock(), but the VM has already started safepoint synchronization, then the thread will block and once it resumes, it will wait on a non-existing round of gc and hangs forever.

k-sareen:
Maybe we can do something similar to what OpenJDK does here [1], that is, temporarily transition to the `_thread_in_vm` state so that the mutator doesn't get blocked on the lock. It might work (unless I've misunderstood the comment and above issue). 

\[1]: https://github.com/mmtk/openjdk/blob/ca90b43f0f51d9ddf754e6ab134c5030cf54118b/src/hotspot/share/runtime/safepoint.cpp#L841-L843

PR #161, opened 2022-06-07
Fix resuming mutators.

wks:
There was bug that if a mutator calls `block_for_gc` after the GC
triggered stop-the-world, it will be block on MMTkHeap::_gc_lock before
reading mmtk_start_the_world_counter, and it will be waiting for a
non-existing GC in the future if the program ends soon.

We now read the mmtk_start_the_world_counter using atomic memory
operations before acquiring MMTkHeap::_gc_lock.  This avoids being
blocked by stop-the-world.

Fixes: https://github.com/mmtk/mmtk-openjdk/issues/117

### DIFF
--- a/openjdk/mmtkUpcalls.cpp
+++ b/openjdk/mmtkUpcalls.cpp
@@ -33,6 +33,7 @@
 #include "mmtkRootsClosure.hpp"
 #include "mmtkUpcalls.hpp"
 #include "mmtkVMCompanionThread.hpp"
+#include "runtime/atomic.hpp"
 #include "runtime/mutexLocker.hpp"
 #include "runtime/os.hpp"
 #include "runtime/safepoint.hpp"
@@ -41,7 +42,8 @@
 #include "runtime/vmThread.hpp"
 #include "utilities/debug.hpp"
 
-static size_t mmtk_start_the_world_count = 0;
+// Note: This counter must be accessed using the Atomic class.
+static volatile size_t mmtk_start_the_world_count = 0;
 
 static void mmtk_stop_all_mutators(void *tls, void (*create_stack_scan_work)(void* mutator)) {
   MMTkHeap::_create_stack_scan_work = create_stack_scan_work;
@@ -79,9 +81,11 @@ static void mmtk_resume_mutators(void *tls) {
   MMTkHeap::heap()->companion_thread()->request(MMTkVMCompanionThread::_threads_resumed, true);
   log_debug(gc)("Mutators resumed. Now notify any mutators waiting for GC to finish...");
 
+  // Note: we don't have to hold gc_lock to increment the counter.
+  Atomic::inc(&mmtk_start_the_world_count);
+
   {
     MutexLockerEx locker(MMTkHeap::heap()->gc_lock(), true);
-    mmtk_start_the_world_count++;
     MMTkHeap::heap()->gc_lock()->notify_all();
   }
   log_debug(gc)("Mutators notified.");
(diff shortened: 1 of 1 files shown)


openjdk/mmtkUpcalls.cpp
```diff
@@ -120,12 +124,25 @@ static void mmtk_spawn_collector_thread(void* tls, int kind, void* ctx) {
...
+
+  // We must read the counter before entering safepoint.
+  // This thread has just triggered GC.
+  // Before this thread enters safe point, the GC cannot start, and therefore cannot finish,
+  // and cannot increment the counter mmtk_start_the_world_count.
+  // Otherwise, if we attempt to acquire the gc_lock first, GC may have triggered stop-the-world
+  // first, and this thread will be blocked for the entire stop-the-world duration before it can
+  // get the lock.  Once that happens, the current thread will read the counter after the GC, and
+  // wait for the next non-existing GC forever.
+  size_t my_count = Atomic::load(&mmtk_start_the_world_count);
```
qinsoon, line 136: The following is just speculation. Let me know if I am wrong.

---

`block_for_gc()` does not mean this thread triggered GC -- it means the thread tries to allocate and MMTk cannot fulfill the allocation request and need the thread to be blocked until a GC is done. 

So `block_for_gc()` may be called multiple times for a GC. The first thread that polls and triggers the GC will be blocked by `block_for_gc()`, but any other thread may also be blocked by `block_for_gc()` if `stop_all_mutators()` does not yet stop that thread, and that thread tries to slow-path allocate. So it is possible that `mmtk_start_the_world_count` is increased multiple times per GC, and that means a thread may be blocked for multiple GCs.

For example, if we have 3 mutator threads, A, B and C, in the worst case:
* thread A in `block_for_gc()`: count = 1
* thread B in `block_for_gc()`: count = 2
* thread C in `block_for_gc()`: count = 3
* GC#1 happens. A resumes, but B and C is still blocked
* thread A in `block_for_gc()`: count = 4
* GC#2 happens. B resumes, but A and C is still blocked.
* ...

Fortunately in this worst case, we will always have 1 thread to resume after each GC. So it won't hang. But this probably will affect performance.

k-sareen: Why not do something like a semaphore where after GC#1, the count is reduced by 1 (or reset to 0)?

wks:
@qinsoon 

> block_for_gc() does not mean this thread triggered GC -- it means the thread tries to allocate and MMTk cannot fulfill the allocation request and need the thread to be blocked until a GC is done.

The second part is right: If more than one threads try to allocate, but fail in the slow path, all of them will call `gc_requester.request()`, and all of them will call `block_for_gc()`, and block until GC is done.  But regarding what "triggering GC" mean, actually `gc_requester.request()` is called before every call to `block_for_gc()`, including both the case of (1) allocation failure (in `Plan::poll`) and (2) `System.gc()` (in `BasePlan::handle_user_collection_request`).  If `gc_requester.request()` triggers GC, then all threads that called `block_for_gc()` must had triggered GC before.

> So it is possible that mmtk_start_the_world_count is increased multiple times per GC

No. This is impossible.  Mutators never increment the counter.  They only read from the counter and see if it is incremented.  Only the GC coordinator thread can increment the counter (calling `resume_mutators` in the `EndOfGC` work packet).

qinsoon (approved):
LGTM

openjdk/mmtkUpcalls.cpp
wks: @k-sareen Maybe, but the idea that mutators can (or need to) modify the synchronising object (the counter, the semaphore, etc.) makes me feel unease, just like the case of "counter incremented multiple times in a GC" that Yi mentioned.

PR #163, opened 2022-06-21
No ProcessEdgesWork in API functions

wks:
Update for upstream PR https://github.com/mmtk/mmtk-core/pull/611

Notably, mmtk-openjdk no longer mentions `ProcessEdgesWork`.

### DIFF
--- a/mmtk/Cargo.toml
+++ b/mmtk/Cargo.toml
@@ -28,9 +28,10 @@ once_cell = "1.10.0"
 # - change branch
 # - change repo name
 # But other changes including adding/removing whitespaces in commented lines may break the CI.
-mmtk = { git = "https://github.com/mmtk/mmtk-core.git", rev = "f78de4ed8ffd6f834c6c3d3a0bde0a651cc33f2b" }
+#mmtk = { git = "https://github.com/mmtk/mmtk-core.git", rev = "f78de4ed8ffd6f834c6c3d3a0bde0a651cc33f2b" }
 # Uncomment the following to build locally
 # mmtk = { path = "../repos/mmtk-core" }
+mmtk = { git = "https://github.com/wks/mmtk-core.git", branch = "api-no-process-edges-work" }
 
 [features]
 default = []
--- a/mmtk/Cargo.toml
+++ b/mmtk/Cargo.toml
@@ -28,10 +28,9 @@ once_cell = "1.10.0"
 # - change branch
 # - change repo name
 # But other changes including adding/removing whitespaces in commented lines may break the CI.
-#mmtk = { git = "https://github.com/mmtk/mmtk-core.git", rev = "571a857b75e648ca4c711d56fbf400d99173398f" }
+mmtk = { git = "https://github.com/mmtk/mmtk-core.git", rev = "f0fed5fc01438162115120c7551d4eb96e888bc2" }
 # Uncomment the following to build locally
 # mmtk = { path = "../repos/mmtk-core" }
-mmtk = { git = "https://github.com/wks/mmtk-core.git", branch = "api-no-process-edges-work" }
 
 [features]
 default = []
--- a/mmtk/src/collection.rs
+++ b/mmtk/src/collection.rs
@@ -1,23 +1,33 @@
-use mmtk::scheduler::WorkBucketStage;
-use mmtk::scheduler::{ProcessEdgesWork, ScanStackRoot};
 use mmtk::util::alloc::AllocationError;
 use mmtk::util::opaque_pointer::*;
 use mmtk::vm::{Collection, GCThreadContext, Scanning, VMBinding};
 use mmtk::{Mutator, MutatorContext};
 
-use crate::OpenJDK;
-use crate::{SINGLETON, UPCALLS};
+use crate::UPCALLS;
+use crate::{MutatorClosure, OpenJDK};
 
 pub struct VMCollection {}
 
-extern "C" fn create_mutator_scan_work<E: ProcessEdgesWork<VM = OpenJDK>>(
-    mutator: &'static mut Mutator<OpenJDK>,
-) {
-    mmtk::memory_manager::add_work_packet(
-        &SINGLETON,
-        WorkBucketStage::Prepare,
-        ScanStackRoot::<E>(mutator),
-    );
+struct ScopedDynamicMutatorVisitorInvoker<'c> {
+    pub closure: &'c mut dyn FnMut(&'static mut Mutator<OpenJDK>),
+}
+
+impl<'c> ScopedDynamicMutatorVisitorInvoker<'c> {
+    fn invoke(&mut self, mutator: &'static mut Mutator<OpenJDK>) {
+        (self.closure)(mutator);
+    }
+
+    extern "C" fn c_invoker(mutator: *mut Mutator<OpenJDK>, me: *mut libc::c_void) {
+        let me: &mut Self = unsafe { &mut *(me as *mut ScopedDynamicMutatorVisitorInvoker) };
+        me.invoke(unsafe { &mut *mutator });
+    }
+
+    fn as_closure(&mut self) -> MutatorClosure {
+        MutatorClosure {
+            func: Self::c_invoker as *const _,
+            data: unsafe { std::mem::transmute(self) },
+        }
+    }
 }
 
 const GC_THREAD_KIND_CONTROLLER: libc::c_int = 0;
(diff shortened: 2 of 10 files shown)


qinsoon (changes requested):
I feel `ScopedDynamicMutatorVisitorInvoker` and `ScopedDynamicFactoryInvoker` are unnecessarily complex. There are a bunch of confusing things happening for those types, and those merely trick the compiler so it won't warn us about unsafe FFI types (if I understand correctly).

I think the issue is the type parameter `<F>` for `stop_all_mutators()`, and the `dyn` pointer `Box<dyn RootsWorkFactory>` for the scan methods. For those, the simplest thing we need is:
1. to pass the value as an opaque pointer to native code
2. to create a method in Rust with `extern "C"` that takes the opaque pointer and invoke the Rust method correctly.

Though those `dyn` pointers are not FFI-safe, I think we fairly safely assume they are just fat opaque pointers. We actually do have such assumptions for our allocator layout already. 

With this assumption (including suppressing such compiler warnings), we should be able to implement the binding side in a more succinct way. However, if we can't, we should rethink about the API change (i.e. the mmtk-core PR). Most of our bindings are written in native languages, and our API should be friendly enough for them.

This issue also exists for the JikesRVM PR: https://github.com/mmtk/mmtk-jikesrvm/pull/114.

wks:
> I feel `ScopedDynamicMutatorVisitorInvoker` and `ScopedDynamicFactoryInvoker` are unnecessarily complex. There are a bunch of confusing things happening for those types, and those merely trick the compiler so it won't warn us about unsafe FFI types (if I understand correctly).

It is a converter.  It converts the rust-style `&dyn` to a C-style closure, i.e. a function pointer + a data pointer.  What is passed to C/C++ is the return value of `as_closure`, i.e. an `EdgesClosure` struct.  Note that `EdgesClosure` has `#[repr(C)]`.  It doesn't contain any `&dyn`.  It contains a `*mut ScopedDynamicFactoryInvoker`, but it is an ordinary pointer, and has the same layout as `void*` in C.

One more difference is that previously, we pass only a function pointer; but now we pass a function pointer + data pointer. This is because the "factory" is an object, and its methods are instance methods instead of top-level functions.  So the data pointer is the counterpart of the "this" pointer.  It is a common practice to pass data pointer alongside function pointer in call-back APIs.  One real-world example is the 

> 
> I think the issue is the type parameter `<F>` for `stop_all_mutators()`, and the `dyn` pointer `Box<dyn RootsWorkFactory>` for the scan methods. For those, the simplest thing we need is:
> 
>     1. to pass the value as an opaque pointer to native code
> 
>     2. to create a method in Rust with `extern "C"` that takes the opaque pointer and invoke the Rust method correctly.
> 

Yes.  It is a valid alternative.

-   Instead of passing function + data, we can simply pass the data (a pointer to the `ScopedDynamicFactoryInvoker` which is basically a wrapper of `&dyn RootsWorkFactory`),
-   We expose its `report_edges_and_renew_buffer` method as an API function (such as `mmtk_report_edges_and_renew_buffer`), so C/C++ can call `mmtk_report_edges_and_renew_buffer` directly instead of calling the function pointer in `EdgesClosure`, which is always the same.

This may be clearer because it passes only an opaque value pointer, and no function pointer.

To make this alternative possible, the factory must be `&dyn`.  If the factory is `&dyn`, we can expose only one single non-generic `mmtk_report_edges_and_renew_buffer` to C.

However, in https://github.com/mmtk/mmtk-core/pull/611, I also mentioned that the root-scanning APIs can have `impl RootsWorkFactory` instead of `dyn RootsWorkFactory` in order to reduce heap allocation. If the factory is `impl` instead of `&dyn`, then all functions that call methods of `impl RootsWorkFactory` must be generic.  Generic Rust functions cannot be exposed to C unless specialised (with applied generic parameters).  It is like our old `fn create_mutator_scan_work<E: ProcessEdgesWork<VM = OpenJDK>>` generic function. We have to provide the `E` parameter to make it a concrete function, so that we can pass it to C as a function pointer.

> 
> Though those `dyn` pointers are not FFI-safe, I think we fairly safely assume they are just fat opaque pointers. We actually do have such assumptions for our allocator layout already.

No. I don't want to make such an assumption.  It is an implementation detail of the Rust compiler.

I see that `BumpAllocator`, `MallocAllocator` and `ImmixAllocator` contain fields of `&dyn` types.  But those fields are at the end of the struct, and C doesn't access them.  In C, we can define a struct which is a *prefix* of the Rust struct, so C only accesses the fields with well-known layout.

> 
> With this assumption (including suppressing such compiler warnings), we should be able to implement the binding side in a more succinct way. However, if we can't, we should rethink about the API change (i.e. the mmtk-core PR). Most of our bindings are written in native languages, and our API should be friendly enough for them.

I prefer safety over succinctness.  The `ScopedDynamicMutatorVisitorInvoker` and the `ScopedDynamicFactoryInvoker` are designed to simplify the implementation of C-style closures by encapsulating details and making it look like an adapter (i.e. wrap it around the factory, and you get a C-style closure).  There is a third-part crate that does similar things, but only works for functions with certain signatures.

Of course there is still room for improvement.  I'll try to make it more like an adapter.

> 
> This issue also exists for the JikesRVM PR: [mmtk/mmtk-jikesrvm#114](https://github.com/mmtk/mmtk-jikesrvm/pull/114).

qinsoon:
> However, in [mmtk/mmtk-core#611](https://github.com/mmtk/mmtk-core/pull/611), I also mentioned that the root-scanning APIs can have `impl RootsWorkFactory` instead of `dyn RootsWorkFactory` in order to reduce heap allocation. If the factory is `impl` instead of `&dyn`, then all functions that call methods of `impl RootsWorkFactory` must be generic. Generic Rust functions cannot be exposed to C unless specialised (with applied generic parameters). It is like our old `fn create_mutator_scan_work<E: ProcessEdgesWork<VM = OpenJDK>>` generic function. We have to provide the `E` parameter to make it a concrete function, so that we can pass it to C as a function pointer.

The heap allocation involved is probably trivial. I don't think there is a need to prematurely optimize that.

> > Though those `dyn` pointers are not FFI-safe, I think we fairly safely assume they are just fat opaque pointers. We actually do have such assumptions for our allocator layout already.
> 
> No. I don't want to make such an assumption. It is an implementation detail of the Rust compiler.
> 
> I see that `BumpAllocator`, `MallocAllocator` and `ImmixAllocator` contain fields of `&dyn` types. But those fields are at the end of the struct, and C doesn't access them. In C, we can define a struct which is a _prefix_ of the Rust struct, so C only accesses the fields with well-known layout.

The thing is that we already assume what `dyn` pointer is:
https://github.com/mmtk/mmtk-openjdk/blob/285f2a9d6822bfe7fc5c2d50a6ce8c75d147dd9b/openjdk/mmtkMutator.hpp#L16-L19

We could just keep using it. If that assumption breaks, we fix it and that will fix all its uses.

> 
> > With this assumption (including suppressing such compiler warnings), we should be able to implement the binding side in a more succinct way. However, if we can't, we should rethink about the API change (i.e. the mmtk-core PR). Most of our bindings are written in native languages, and our API should be friendly enough for them.
> 
> I prefer safety over succinctness. The `ScopedDynamicMutatorVisitorInvoker` and the `ScopedDynamicFactoryInvoker` are designed to simplify the implementation of C-style closures by encapsulating details and making it look like an adapter (i.e. wrap it around the factory, and you get a C-style closure). There is a third-part crate that does similar things, but only works for functions with certain signatures.

I don't really see the converter is safer. For example, the `MutatorClosure` you create from `struct ScopedDynamicMutatorVisitorInvoker<'c>` contains a pointer to the struct, but it does not have any lifetime bound to it. This is just one thing that I spotted when reviewing the code. The complexity in the converter makes it hard to reason that it is actually safe.

wks:
I eliminated the `ScopedDynamicXxxxxInvoker` and the use of `dyn`.  mmtk-core now passes an `impl RootsWorkFactory` to the binding.  So the binding now just pass a pointer to the `RootsWorkFactory` instance to C++, and C++ will pass it back.

However, since it is `impl`, the call-back function needs to be specialised.  Like the old code `create_process_edges_work::<W>`, we pass a type-parameterised function `report_edges_and_renew_buffer::<F>` to C as a call-back.

Now the implementations of `RootsWorkFactory` (in mmtk-core) decide whether to use `Arc` or `Box` inside. It is now an implementation detail, and not part of the API.

qinsoon (approved):
LGTM

PR #165, opened 2022-06-28
Fix root scanning

wenyuzhao:
* Remove incorrect `MMTkRootScanWorkScope`
* Optimize `CodeCache` roots scanning
* Remove unused `compute_*_roots` functions

# Performance comparison

_Machine: Shrew (Core i9-9900K Coffee Lake, 8/16 cores)_

### Immix

http://squirrel.anu.edu.au/plotty-public/wenyuz/v8/p/3k2Ucr

Screenshot 2022-06-30 at 11 03 12 am

### MarkCompact

http://squirrel.anu.edu.au/plotty-public/wenyuz/v8/p/WjS3eX

Screenshot 2022-06-30 at 11 03 49 am

### DIFF
--- a/mmtk/Cargo.toml
+++ b/mmtk/Cargo.toml
@@ -22,6 +22,7 @@ openjdk_version = "ca90b43f0f51d9ddf754e6ab134c5030cf54118b"
 libc = "0.2"
 lazy_static = "1.1"
 once_cell = "1.10.0"
+spin = "0.9.2"
 # Be very careful to commit any changes to the following mmtk dependency, as our CI scripts (including mmtk-core CI)
 # rely on matching these lines to modify them: e.g. comment out the git dependency and use the local path.
 # These changes are safe:
--- a/mmtk/Cargo.toml
+++ b/mmtk/Cargo.toml
@@ -22,7 +22,6 @@ openjdk_version = "ca90b43f0f51d9ddf754e6ab134c5030cf54118b"
 libc = "0.2"
 lazy_static = "1.1"
 once_cell = "1.10.0"
-spin = "0.9.2"
 # Be very careful to commit any changes to the following mmtk dependency, as our CI scripts (including mmtk-core CI)
 # rely on matching these lines to modify them: e.g. comment out the git dependency and use the local path.
 # These changes are safe:
--- a/mmtk/src/api.rs
+++ b/mmtk/src/api.rs
@@ -15,7 +15,9 @@ use mmtk::Mutator;
 use mmtk::MutatorContext;
 use mmtk::MMTK;
 use once_cell::sync;
+use std::cell::RefCell;
 use std::ffi::{CStr, CString};
+use std::sync::atomic::Ordering;
 
 // Supported barriers:
 static NO_BARRIER: sync::Lazy<CString> = sync::Lazy::new(|| CString::new("NoBarrier").unwrap());
@@ -280,5 +282,37 @@ pub extern "C" fn get_finalized_object() -> ObjectReference {
     match memory_manager::get_finalized_object(&SINGLETON) {
         Some(obj) => obj,
         None => unsafe { Address::ZERO.to_object_reference() },
+    }
+}
+
+thread_local! {
+    static NMETHOD_SLOTS: RefCell<Vec<Address>> = RefCell::new(vec![]);
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn mmtk_add_nmethod_oop(addr: Address) {
+    NMETHOD_SLOTS.with(|x| x.borrow_mut().push(addr))
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn mmtk_register_nmethod(nm: Address) {
+    let slots = NMETHOD_SLOTS.with(|x| {
+        if x.borrow().len() == 0 {
+            return None;
+        }
+        Some(x.replace(vec![]))
+    });
+    if slots.is_none() {
+        return;
+    }
+    let slots = slots.unwrap();
+    crate::CODE_CACHE_ROOTS_SIZE.fetch_add(slots.len(), Ordering::SeqCst);
+    crate::CODE_CACHE_ROOTS.lock().unwrap().insert(nm, slots);
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn mmtk_unregister_nmethod(nm: Address) {
+    if let Some(slots) = crate::CODE_CACHE_ROOTS.lock().unwrap().remove(&nm) {
+        crate::CODE_CACHE_ROOTS_SIZE.fetch_sub(slots.len(), Ordering::SeqCst);
     }
 }
--- a/mmtk/src/api.rs
+++ b/mmtk/src/api.rs
@@ -286,14 +286,19 @@ pub extern "C" fn get_finalized_object() -> ObjectReference {
 }
 
 thread_local! {
+    /// Cache all the pointers reported by the current thread.
     static NMETHOD_SLOTS: RefCell<Vec<Address>> = RefCell::new(vec![]);
 }
 
+/// Report a list of pointers in nmethod to mmtk.
 #[no_mangle]
 pub extern "C" fn mmtk_add_nmethod_oop(addr: Address) {
     NMETHOD_SLOTS.with(|x| x.borrow_mut().push(addr))
 }
 
+/// Register a nmethod.
+/// The c++ part of the binding should scan the nmethod and report all the pointers to mmtk first, before calling this function.
+/// This function will transfer all the locally cached pointers of this nmethod to the global storage.
 #[no_mangle]
 pub extern "C" fn mmtk_register_nmethod(nm: Address) {
     let slots = NMETHOD_SLOTS.with(|x| {
@@ -302,17 +307,20 @@ pub extern "C" fn mmtk_register_nmethod(nm: Address) {
         }
         Some(x.replace(vec![]))
     });
-    if slots.is_none() {
-        return;
-    }
-    let slots = slots.unwrap();
-    crate::CODE_CACHE_ROOTS_SIZE.fetch_add(slots.len(), Ordering::SeqCst);
-    crate::CODE_CACHE_ROOTS.lock().unwrap().insert(nm, slots);
+    let slots = match slots {
+        Some(slots) => slots,
+        _ => return,
+    };
+    let mut roots = crate::CODE_CACHE_ROOTS.lock().unwrap();
+    crate::CODE_CACHE_ROOTS_SIZE.store(crate::CODE_CACHE_ROOTS_SIZE.load(Ordering::Relaxed) + slots.len(), Ordering::Relaxed);
+    roots.insert(nm, slots);
 }
 
+/// Unregister a nmethod.
 #[no_mangle]
 pub extern "C" fn mmtk_unregister_nmethod(nm: Address) {
-    if let Some(slots) = crate::CODE_CACHE_ROOTS.lock().unwrap().remove(&nm) {
-        crate::CODE_CACHE_ROOTS_SIZE.fetch_sub(slots.len(), Ordering::SeqCst);
+    let mut roots = crate::CODE_CACHE_ROOTS.lock().unwrap();
+    if let Some(slots) = roots.remove(&nm) {
+        crate::CODE_CACHE_ROOTS_SIZE.store(crate::CODE_CACHE_ROOTS_SIZE.load(Ordering::Relaxed) - slots.len(), Ordering::Relaxed);
     }
 }
--- a/mmtk/src/api.rs
+++ b/mmtk/src/api.rs
@@ -312,6 +312,7 @@ pub extern "C" fn mmtk_register_nmethod(nm: Address) {
         _ => return,
     };
     let mut roots = crate::CODE_CACHE_ROOTS.lock().unwrap();
+    // Relaxed add instead of `fetch_add`, since we've already acquired the lock.
     crate::CODE_CACHE_ROOTS_SIZE.store(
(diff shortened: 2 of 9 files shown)


openjdk/mmtkUpcalls.cpp
```diff
@@ -65,10 +65,12 @@ static void mmtk_stop_all_mutators(void *tls, void (*create_stack_scan_work)(voi
     }
   }
   log_debug(gc)("Finished enumerating threads.");
+  nmethod::oops_do_marking_prologue();
 }
 
 static void mmtk_resume_mutators(void *tls) {
-  ClassLoaderDataGraph::purge();
+  nmethod::oops_do_marking_epilogue();
```
k-sareen, line 72: Is is correct to call the epilogue after the GC instead of after root scanning? I honestly don't have the knowledge to understand what the functions are doing, so am unsure if it is considered correct/okay to call the epilogue after the GC has occurred.

openjdk/mmtkHeap.hpp
```diff
@@ -181,9 +181,11 @@ class MMTkHeap : public CollectedHeap {
 
   // An object is scavengable if its location may move during a scavenge.
   // (A scavenge is a GC which is not a full GC.)
-  bool is_scavengable(oop obj);
+  inline bool is_scavengable(oop obj) { return true; }
```
k-sareen, line 184: This is only true for generational GCs right? Other full heap GCs like `MarkCompact` won't be scavenging

wenyuzhao: I'm not sure as well. But from my experience, any moving GC in mmtk should return true here, otherwise it may crash for some reason...

openjdk/mmtkUpcalls.cpp
wenyuzhao: Looks like this function only prunes an internal nmethod list. I guess it should work correctly as long as all the necessary nmethods are scanned properly.

openjdk/mmtkHeap.hpp
k-sareen: Seems like ZGC returns `false` [1] while Shenandoah returns `true` [2]. Other GCs will conditionally return `true` if the object is in the nursery. Maybe we should expose a `is_movable_in_current_gc()` from mmtk-core that answers this question instead of unconditionally returning `true`.

Unsure why ZGC returns `false` though. I guess it technically isn't generational. But then why does Shenandoah return `true` without being generational. Does Shenandoah do incremental collections?

\[1]: https://github.com/mmtk/openjdk/blob/jdk-11.0.15%2B8-mmtk/src/hotspot/share/gc/z/zCollectedHeap.cpp#L109-L111
\[2]: https://github.com/mmtk/openjdk/blob/jdk-11.0.15%2B8-mmtk/src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp#L509-L510

qinsoon: I think normally if you state a GC will move objects but actually it won't move objects, it is more likely to be a performance issue rather than a correctness issue. If a GC is not moving, the runtime can make some assumptions about it to improve performance.

But it does not seem to be the case: before the change, we tell the runtime that we won't move objects in a 'scavenge GC', but we actually move objects in quite a few GC plans. I am wondering if this function is actually used or not. If we does not even tell the runtime that we are doing 'scavenge GC', then this function is probably irrelevant.

I am wondering what is the reason for changing the function.

k-sareen: It seems like the `is_scavengable()` function is primarily used to check if an oop can be embedded in code (if it is scavengable, then it can't be embedded in code) and for the `do_oop()` visitor for the `DetectScavengableRoot` oop closure for nmethods.

k-sareen:
Wow! Nice work! 40% improvement on time.stw for `Immix`. Weird that `MarkCompact` is a bit slower for some benchmarks though. I've also noticed that this PR is quite similar to #141 as well. Should we consolidate them or try to merge #141 after this one? I will be able to properly test this PR after I'm done with my MPLR deadline (actual deadline is 15th July, but ideally I'm done before ~10th). I apologize for the delay.

openjdk/mmtkHeap.hpp
wenyuzhao: Agreed. However, the change here is simply adding an `inline` attribute so that the function body can be inlined. We can switch to a more correct implementation in the future.

wenyuzhao:
> Wow! Nice work! 40% improvement on time.stw for `Immix`. Weird that `MarkCompact` is a bit slower for some benchmarks though. I've also noticed that this PR is quite similar to #141 as well. Should we consolidate them or try to merge #141 after this one? I will be able to properly test this PR after I'm done with my MPLR deadline (actual deadline is 15th July, but ideally I'm done before ~10th). I apologize for the delay.

I think the two PRs are different, although they share some common changes to the CodeCache. This PR tries to optimize CodeCache scanning and does not affect correctness.  #141 tries to fix memory leak caused by incorrect root scanning.  Also when I was working on #141, I haven't figured out how to scan CodeCache efficiently.

After this PR, I will continue working on #141. It still has some difficulty with MarkCompact at the moment...

mmtk/src/lib.rs
```diff
@@ -128,3 +129,7 @@ lazy_static! {
 #[no_mangle]
 pub static MMTK_MARK_COMPACT_HEADER_RESERVED_IN_BYTES: usize =
     mmtk::util::alloc::MarkCompactAllocator::<OpenJDK>::HEADER_RESERVED_IN_BYTES;
+
+static CODE_CACHE_ROOTS: spin::Lazy<Mutex<HashMap<Address, Vec<Address>>>> =
+    spin::Lazy::new(|| Mutex::new(HashMap::new()));
+static CODE_CACHE_ROOTS_SIZE: AtomicUsize = AtomicUsize::new(0);
```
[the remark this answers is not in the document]
qinsoon: Also in `mmtk_register_nmethod()`, you modify the `SIZE` before acquiring a lock to the `ROOTS`, which will make the `SIZE` and the `ROOTS` inconsistent.

qinsoon (changes requested):
Very impressive performance improvement. I have some questions about the code, especially about the thread local `NMETHOD_SLOTS`. Also with the merging of https://github.com/mmtk/mmtk-openjdk/pull/163, this PR will need to adapt the changes.

mmtk/src/lib.rs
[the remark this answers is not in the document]
wenyuzhao: Yes, `SIZE` is not necessary, it's just an optimization. Because `CODE_CACHE_ROOTS` is a set of `Vec<Address>`s, at root scanning time these vectors have to be merged together as a single buffer.  Record total `SIZE` ahead of time and use it later for the merged buffer allocation may prevent frequent vector size expansion.

The consistency is ensured by separating the writes and reads. `SIZE` is only updated during mutator phases, and reads only happen in pauses.

---

Apart from tracking `SIZE`, I also tried putting each vector in the cached roots (`HashMap<Address, Vec<Address>>`) to a different work packet.   There's a slight performance drop by doing so, since this may yield too many vector clones, too many work packets, and most of the vectors are small.

mmtk/src/lib.rs
```diff
@@ -128,3 +129,7 @@ lazy_static! {
 #[no_mangle]
 pub static MMTK_MARK_COMPACT_HEADER_RESERVED_IN_BYTES: usize =
     mmtk::util::alloc::MarkCompactAllocator::<OpenJDK>::HEADER_RESERVED_IN_BYTES;
+
+static CODE_CACHE_ROOTS: spin::Lazy<Mutex<HashMap<Address, Vec<Address>>>> =
```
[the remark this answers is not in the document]
wenyuzhao: Sure I can switch to `lazy_static`. I just prefer not using macros:)

mmtk/src/api.rs
```diff
@@ -282,3 +284,35 @@ pub extern "C" fn get_finalized_object() -> ObjectReference {
         None => unsafe { Address::ZERO.to_object_reference() },
     }
 }
+
+thread_local! {
```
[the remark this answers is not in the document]
wenyuzhao: Sure I will add comments explaining this.

Basically, when the OpenJDK heap registers a `nmethod`, the c++ part will scan all the object pointers in the `nmethod`, and call `mmtk_add_nmethod_oop` once for each reported pointer. The rust part will record the pointers in a thread-local variable.

After the scanning is done, the c++ part will call `mmtk_register_nmethod` to move the pointers recorded in the thread-local variable to the global storage (`CODE_CACHE_ROOTS`).

openjdk/mmtkHeap.cpp
```diff
@@ -335,135 +335,74 @@ void MMTkHeap::print_tracing_info() const {
...
   }
+public:
+  void do_oop(oop* p)       { do_oop_work(p); }
+  void do_oop(narrowOop* p) { do_oop_work(p); }
 };
 
-void MMTkHeap::scan_static_roots(OopClosure& cl) {
+
+void MMTkHeap::register_nmethod(nmethod* nm) {
+  MMTkRegisterNMethodOopClosure reg_cl;
```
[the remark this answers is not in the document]
wenyuzhao: `MMTkRegisterNMethodOopClosure` is used to scan the pointers in the `nmethod`.  The rust part will push the pointers to a thread-local vector first, then move to the global storage, to reduce lock contentions.

mmtk/src/lib.rs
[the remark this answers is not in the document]
qinsoon: > Yes, `SIZE` is not necessary, it's just an optimization. Because `CODE_CACHE_ROOTS` is a set of `Vec<Address>`s, at root scanning time these vectors have to be merged together as a single buffer. Record total `SIZE` ahead of time and use it later for the merged buffer allocation may prevent frequent vector size expansion.

You can just calculate the total size of the `Vec<Address>` before creating the vec.
```rust
let code_cache_roots = CODE_CACHE_ROOTS.lock().unwrap();
let size = code_cache_roots.values().map(|vec| vec.len()).sum();
let mut vec = Vec::with_capacity(size);
for roots in code_cache_roots.values() {
    for r in roots {
        vec.push(*r)
    }
}
```

My point is that if you have acquired the lock to `CODE_CACHE_ROOTS` whenever you access `CODE_CACHE_ROOTS_SIZE`, then there is little point having a separate count.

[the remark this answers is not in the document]
wenyuzhao: I'm afraid calculating the size in a loop may slow down the GC time. If I can calculate the total size cheaply ahead of time, simply using the pre-calculated size here is the fastest way.

This makes me realize that if I move the `SIZE` increments after the lock, I can do relaxed instead of atomic increments. This solves the consistency issue, has higher the increments performance, while still having a precise pre-calculated size counter.

mmtk/src/api.rs
```diff
@@ -282,3 +284,49 @@ pub extern "C" fn get_finalized_object() -> ObjectReference {
...
+            return None;
+        }
+        Some(x.replace(vec![]))
+    });
+    let slots = match slots {
+        Some(slots) => slots,
+        _ => return,
+    };
+    let mut roots = crate::CODE_CACHE_ROOTS.lock().unwrap();
+    crate::CODE_CACHE_ROOTS_SIZE.store(
```
k-sareen, line 315: I'm unsure why you are using a load+store instead of a `fetch_add()` here (and `fetch_sub()` for unregister nmethod)

k-sareen: Sure. But I thought `fetch_add(number, Ordering::Relaxed)` would be the same? Or at least as per my understanding of the documentation for `fetch_add()` [1]

\[1]: https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.fetch_add

qinsoon: Yeah. `fetch_add` will result in a `lock add` instruction, and load-store will result in a simple `add`. 

However, for cases like this, the most intuitive way is to just use `fetch_add()`. If you are doing something that is different, it would be better to put some comments to it so people understand and won't attempt to 'fix/refactor' it.

qinsoon (approved):
LGTM

mmtk/src/api.rs
wenyuzhao: Sure I'll add a comment to this

mmtk/Cargo.toml
```diff
@@ -22,6 +22,7 @@ openjdk_version = "ca90b43f0f51d9ddf754e6ab134c5030cf54118b"
 libc = "0.2"
 lazy_static = "1.1"
 once_cell = "1.10.0"
+spin = "0.9.2"
```
k-sareen, line 25: If we're not using `spin` (it seems like we're not) then we should remove it.

wenyuzhao: Thanks! Removed

mmtk/src/lib.rs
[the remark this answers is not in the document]
wks: If you always access the `CODE_CACHE_ROOTS_SIZE` while holding the lock of `CODE_CACHE_ROOTS`, you can merge them into one `Mutex<T>`, like this:

```rust
struct CodeCacheRoots {
    hash_map: HashMap<Address, Vec<Address>>,
    size: usize,
}

lazy_static! {
    static ref CODE_CACHE_ROOTS: Mutex<CodeCacheRoots> = Mutex::new(CodeCacheRoots {
        hash_map: HashMap::new(),
        size: 0,
    });
}
```

When using, you lock the mutex and have access to both `hash_map` and `size`:
```rust
{
    let mut lock_guard = CODE_CACHE_ROOTS.lock().unwrap();
    *lock_guard.size += slots.len();
    lock_guard.hash_map.insert(nm, slots);
}
```

PR #167, opened 2022-07-21
Fix build issue when we use plan features

qinsoon:


wks (approved):
LGTM, but we can now set the property directly, like this:
`BUILDER.lock().unwrap().options.plan.set(PlanSelector::Immix);`.

PR #174, opened 2022-08-10
Get build info from mmtk-core

qinsoon:
This PR implements `ThirdPartyHeap::version()` for https://github.com/mmtk/openjdk/pull/14.


wks (approved):
LGTM

k-sareen:
I know the PR has merged already, but can I suggest adding the enabled features to the version string (or somewhere in `--version`) as well? For example, what I do when I copy builds is I name the build as `jdk11-core_<mmtk-core git-hash>-jdk_<mmtk-openjdk git-hash>-vm_<openjdk git-hash>-<feature1>-<feature2>`. It would make the builds more reproducible.

PR #175, opened 2022-08-17
Mimalloc Allocator

paigereeves:
This pull request introduces an allocator based on Microsoft's mimalloc (https://www.microsoft.com/en-us/research/uploads/prod/2019/06/mimalloc-tr-v1.pdf) used in a MarkSweep GC. It is largely completed, but it needs some cleaning up and logical rearrangement.


qinsoon:
It seems that I don't have write permission to update this PR (I can update the mmtk-core PR). I pushed the changes in the PR to a local branch, and created a new PR from that: https://github.com/mmtk/mmtk-openjdk/pull/180. I will keep working on this from that PR. I will close this one.

PR #177, opened 2022-08-26
Refactor barrier interface

wenyuzhao:
This PR updates mmtk-openjdk to use the new barrier API.

mmtk-core PR: https://github.com/mmtk/mmtk-core/pull/650


openjdk/mmtkBarrierSetAssembler_x86.hpp
```diff
@@ -4,11 +4,40 @@
...
+  /// Barrier elision test
+  virtual bool can_remove_barrier(DecoratorSet decorators, Register val, bool skip_const_null) const {
+    bool in_heap = (decorators & IN_HEAP) != 0;
+    bool as_normal = (decorators & AS_NORMAL) != 0;
+    assert((decorators & IS_DEST_UNINITIALIZED) == 0, "unsupported");
+    return !in_heap || (skip_const_null && val == noreg);
+  }
+
+  /// Generate C1 write barrier slow-call assembly code
+  virtual void generate_c1_write_barrier_runtime_stub(StubAssembler* sasm) const;
```
qinsoon, line 30: What is the difference between `generate_c1_write_barrier_runtime_stub` and `gen_c1_generic_write_barrier_stub`? They seem to generate similar code that calls the slow path, but used in different cases. You should consider renaming them, as the difference in their names does not tell what they are.

qinsoon (commented):
LGTM. But I don't have much knowledge about OpenJDK. You may want to get an approval from other reviewers.

qinsoon:
@wks @caizixian Can you review this? It would be good if either of you can review it.

openjdk/mmtkBarrierSetC1.hpp
```diff
@@ -4,27 +4,84 @@
...
     BarrierSetC1::store_at_resolved(access, value);
+    if (access.is_oop()) object_reference_write_post(access, access.base().opr(), access.resolved_addr(), value);
   }
 
-  virtual LIR_Opr atomic_cmpxchg_at_resolved(LIRAccess& access, LIRItem& cmp_value, LIRItem& new_value) {
-    return BarrierSetC1::atomic_cmpxchg_at_resolved(access, cmp_value, new_value);
+  /// Substituting write barrier (cmpxchg)
+  virtual LIR_Opr atomic_cmpxchg_at_resolved(LIRAccess& access, LIRItem& cmp_value, LIRItem& new_value) override {
+    if (access.is_oop()) object_reference_write_pre(access, access.base().opr(), access.resolved_addr(), new_value.result());
+    LIR_Opr result = BarrierSetC1::atomic_cmpxchg_at_resolved(access, cmp_value, new_value);
```
wks, line 30: Cmpxchg, when successful, performs both a read and a write operation.  But if cmpxchg fails, the write operation is not performed.  Should we skip the post write barrier if it fails?

What does cmpxchg do in vanilla OpenJDK?

openjdk/mmtkBarrierSetC2.hpp
```diff
@@ -46,17 +46,36 @@ class TypeFunc;
...
-    return BarrierSetC2::store_at_resolved(access, val);
+    if (access.is_oop()) object_reference_write_pre(access.kit(), access.base(), access.addr().node(), val.node());
+    Node* store = BarrierSetC2::store_at_resolved(access, val);
+    if (access.is_oop()) object_reference_write_post(access.kit(), access.base(), access.addr().node(), val.node());
+    return store;
   }
   virtual Node* atomic_cmpxchg_val_at_resolved(C2AtomicAccess& access, Node* expected_val, Node* new_val, const Type* value_type) const {
-    return BarrierSetC2::atomic_cmpxchg_val_at_resolved(access, expected_val, new_val, value_type);
+     if (access.is_oop()) object_reference_write_pre(access.kit(), access.base(), access.addr().node(), new_val);
+    Node* result = BarrierSetC2::atomic_cmpxchg_val_at_resolved(access, expected_val, new_val, value_type);
```
wks, line 64: Similar to C1, should we skip the write barrier if cmpxchg fails?

openjdk/barriers/mmtkObjectBarrier.cpp
```diff
@@ -69,18 +49,32 @@ void MMTkObjectBarrierSetAssembler::record_modified_node(MacroAssembler* masm, R
...
 #endif
 }
 
+void MMTkObjectBarrierSetAssembler::arraycopy_epilogue(MacroAssembler* masm, DecoratorSet decorators, BasicType type, Register src, Register dst, Register count) {
+  const bool dest_uninitialized = (decorators & IS_DEST_UNINITIALIZED) != 0;
+  if ((type == T_OBJECT || type == T_ARRAY) && !dest_uninitialized) {
+    __ pusha();
+    __ movptr(c_rarg0, src);
+    __ movptr(c_rarg1, dst);
+    __ movptr(c_rarg3, count);
```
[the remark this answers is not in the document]
wenyuzhao: Fixed

openjdk/mmtkBarrierSet.cpp
```diff
@@ -130,3 +130,23 @@ void MMTkBarrierSet::print_on(outputStream* st) const {
...
+  ::mmtk_object_reference_write_slow((MMTk_Mutator) &Thread::current()->third_party_heap_mutator, src, slot, target);
+}
+
+void MMTkBarrierSetRuntime::object_reference_array_copy_pre_call(void* src, void* dst, size_t count) {
+  ::mmtk_array_copy_pre((MMTk_Mutator) &Thread::current()->third_party_heap_mutator, src, dst, count);
+}
+
+void MMTkBarrierSetRuntime::object_reference_array_copy_post_call(void* src, void* dst, size_t count) {
+  ::mmtk_array_copy_post((MMTk_Mutator) &Thread::current()->third_party_heap_mutator, src, dst, count);
+}
```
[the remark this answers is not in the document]
wenyuzhao: Fixed

openjdk/mmtkBarrierSet.hpp
```diff
@@ -54,12 +55,37 @@ struct MMTkAllocatorOffsets {
  */
 MMTkAllocatorOffsets get_tlab_top_and_end_offsets(AllocatorSelector selector);
 
+#define FN_ADDR(function) CAST_FROM_FN_PTR(address, function)
```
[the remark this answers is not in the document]
wenyuzhao: It's actually being used at 7 places. So I thought it might be worth to apply this macro

openjdk/mmtkBarrierSetC1.hpp
wenyuzhao: I was following OpenJDK's style, which does not check the result status either. Here're two examples:
* [C1 atomic xchg barrier](https://github.com/mmtk/openjdk/blob/master/src/hotspot/share/gc/shared/c1/modRefBarrierSetC1.cpp#L75-L81)
* [C2 atomic xchg barrier](https://github.com/mmtk/openjdk/blob/master/src/hotspot/share/gc/shared/c2/modRefBarrierSetC2.cpp#L73-L81)

openjdk/mmtkBarrierSetC2.cpp
```diff
@@ -593,3 +593,43 @@ void MMTkBarrierSetC2::expand_allocate(PhaseMacroExpand* x,
...
+
+bool MMTkBarrierSetC2::can_remove_barrier(GraphKit* kit, PhaseTransform* phase, Node* src, Node* slot, Node* val, bool skip_const_null) const {
+  // Skip barrier if the new target is a null pointer.
+  if (skip_const_null && val != NULL && val->is_Con() && val->bottom_type() == TypePtr::NULL_PTR) {
+    return true;
+  }
+  // Barrier elision based on allocation node does not working well with slowpath-only allocation.
+  if (!MMTK_ENABLE_ALLOCATION_FASTPATH) return false;
+  // No barrier required for newly allocated objects.
+  if (src == kit->just_allocated_object(kit->control())) return true;
```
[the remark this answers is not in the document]
wenyuzhao: This is generic to all barriers with young optimization. LXR uses this as well. If a barrier wants to track young objects, it can override this method and switch to their own barrier elision policy.

openjdk/mmtkBarrierSetAssembler_x86.hpp
wenyuzhao: `generate_c1_write_barrier_runtime_stub` will generate a small assembly function that calls MMTk barrier slow-calls. It's called a "runtime stub".

`gen_c1_generic_write_barrier_stub` will generate a piece of code in the barrier to jump to the above assembly function.

So in general, to do a slow-path barrier call in C1, it will jump to the "runtime stub" first, then call to the mmtk API.

I directly copied this code from G1, and currently I'm still not sure why jumping to the "runtime stub" first is necessary. Probably in the future I'll see if I can remove this "runtime stub".

qinsoon: I see. `CodeStub` is a piece of out-of-line code (a new function) that is put at the end of a function. There is only one copy of that for each function, and for places you need it, you just generate some code to call it.

I suggest renaming `gen_c1_generic_write_barrier_stub` to `generate_c1_write_barrier_stub_call`, or something like that. It is not really generating a stub.

wenyuzhao: Yes makes sense. Renamed.

openjdk/mmtkBarrierSetC1.hpp
wks: I think it is OK to apply card marking and object remembering barriers conservatively even when the object is not actually changed.  It is OK for now, but I am not sure if other barriers are OK, too.  Let's discuss this separately.

wks (approved):
I don't have further comments. It looks good to me now.

PR #180, opened 2022-09-26
Mimalloc marksweep support

qinsoon:


openjdk/mmtkMutator.cpp
```diff
@@ -34,3 +38,7 @@ HeapWord* MMTkMutatorContext::alloc(size_t bytes, Allocator allocator) {
 void MMTkMutatorContext::flush() {
   ::flush_mutator((MMTk_Mutator) this);
 }
+
+void MMTkMutatorContext::destroy() {
+  ::destroy_mutator((MMTk_Mutator) this);
+}
```
wks, line 44: This file doesn't end with a newline character.  Although C++11 doesn't forbid it, it may cause problems in unexpected ways, especially when we process files using scripts.

openjdk/mmtkMutator.cpp
```diff
@@ -6,6 +6,10 @@
 size_t MMTkMutatorContext::max_non_los_default_alloc_bytes = 0;
 
 MMTkMutatorContext MMTkMutatorContext::bind(::Thread* current) {
+    if (FREE_LIST_ALLOCATOR_SIZE != sizeof(FreeListAllocator)) {
```
wks, line 9: This line is not properly indented.
```suggestion
  if (FREE_LIST_ALLOCATOR_SIZE != sizeof(FreeListAllocator)) {
```

wks (changes requested):
Only some small problems remain.

wks (approved):
LGTM

PR #182, opened 2022-09-27
Make function names consistent between Rust and C++

### LINKED ISSUE #181: Function names differ between Rust and C++ upcalls struct
Some function names such as `spawn_gc_thread` have different names in the Rust [1] and C++ [2] versions of the OpenJDK_Upcalls struct. Should be fixed for consistency.

\[1]: https://github.com/mmtk/mmtk-openjdk/blob/b3307bba3bed0fb71ad7096ee019c9b1ca655ca9/mmtk/src/lib.rs#L63
\[2]: https://github.com/mmtk/mmtk-openjdk/blob/b3307bba3bed0fb71ad7096ee019c9b1ca655ca9/openjdk/mmtk.h#L144

k-sareen:
Closes #181


qinsoon (approved):
LGTM. Thanks for the fix.

PR #187, opened 2022-11-17
Migrate to 2021 edition

wks:
This migrates mmtk-openjdk to Rust 2021 edition, the same as mmtk-core.  It should not change the behaviour of mmtk-openjdk in any way.

Background: I was experimenting with moving some reference-processing mechanism from mmtk-core to mmtk-openjdk, but the fact that mmtk-openjdk is still based on Rust 2015 makes things difficult, mainly because of syntactic incompatibilities.

### DIFF
--- a/mmtk/Cargo.lock
+++ b/mmtk/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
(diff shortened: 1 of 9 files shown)


qinsoon (approved):
LGTM

PR #188, opened 2022-11-25
Fix function pointer type.

wks:
That was a mistake.  I wrote `*const extern "C" fn(...)` and thought that was function pointer type, but it is actually "a pointer to a pointer to a function".  The type of a function pointer to a plain function with C calling convention is just `extern "C" fn(....)`.

p.s. I wonder why this has been working so far.  Maybe the Rust side used an unsafe cast to write the right function pointer to a field with wrong pointer type, and the C++ side's signature is actually correct. So it worked even though the typing was incorrect.

### DIFF
--- a/mmtk/src/collection.rs
+++ b/mmtk/src/collection.rs
@@ -8,11 +8,13 @@ use crate::{MutatorClosure, OpenJDK};
 
 pub struct VMCollection {}
 
-extern "C" fn report_mutator_stop<F>(mutator: *mut Mutator<OpenJDK>, callback: *mut F)
-where
+extern "C" fn report_mutator_stop<F>(
+    mutator: *mut Mutator<OpenJDK>,
+    callback_ptr: *mut libc::c_void,
+) where
     F: FnMut(&'static mut Mutator<OpenJDK>),
 {
-    let callback: &mut F = unsafe { &mut *callback };
+    let callback: &mut F = unsafe { &mut *(callback_ptr as *mut F) };
     callback(unsafe { &mut *mutator });
 }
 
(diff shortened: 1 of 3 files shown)


mmtk/src/collection.rs
```diff
@@ -8,11 +8,13 @@ use crate::{MutatorClosure, OpenJDK};
 
 pub struct VMCollection {}
 
-extern "C" fn report_mutator_stop<F>(mutator: *mut Mutator<OpenJDK>, callback: *mut F)
-where
+extern "C" fn report_mutator_stop<F>(
+    mutator: *mut Mutator<OpenJDK>,
+    callback_ptr: *mut libc::c_void,
```
qinsoon, line 13: Why is this changed from `*mut F` to `*mut c_void`?

wks: Because the `MutatorClosure` struct needs to be passed to C.

```
#[repr(C)]
pub struct MutatorClosure {
    pub func: extern "C" fn(mutator: *mut Mutator<OpenJDK>, data: *mut libc::c_void),
    pub data: *mut libc::c_void,
}
```

The contract here is, C/C++ programs treat `data` as an opaque pointer, and use it as the last argument of `func`.  The `data` may not point to the same thing each time, so it is `*mut libc::c_void`.

Another reason is, Rust forbids the cast between function pointer types with different parameter types, so the line `func: report_mutator_stop::<F> as *const _` which used to work now stopped working.

So the only solution is to let the Rust function `report_mutator_stop<F>` match the parameter type `*mut libc::c_void`.

qinsoon (approved):
LGTM

qinsoon:
Can you merge this after https://github.com/mmtk/mmtk-openjdk/pull/186?

wks:
> Can you merge this after #186?

Sure.

PR #191, opened 2022-12-26
Workaround soft_ref_policy crash

wks:
Sometimes the VM will call methods of CollectedHeap::soft_ref_policy(). If it returns nullptr, the VM will crash for SIGSEGV in a subsequent method invocation on the returned nullptr.

This is just a workaround because our soft reference processor is currently implemented in mmtk-core.


qinsoon (approved):
LGTM

PR #192, opened 2023-01-04
#include arrayOop.hpp in mmtkBarrierSet.hpp

jjoo172:
This is a dependency, and without this include, it causes a build failure with our local build configuration. (Possibly related to `--disable-precompiled-headers`, but haven't confirmed which exact flag causes this issue.)


qinsoon (approved):
LGTM. Thanks for the PR.

PR #193, opened 2023-01-24
Add some tests for StickyImmix

qinsoon:
This PR updates to mmtk-core https://github.com/mmtk/mmtk-core/pull/741, and adds some tests for the new sticky immix plan.


mmtk/src/object_model.rs
```diff
@@ -106,4 +106,9 @@ impl ObjectModel<OpenJDK> for VMObjectModel {
             ((*UPCALLS).dump_object)(object);
         }
     }
+
+    fn is_object_sane(object: ObjectReference) -> bool {
```
k-sareen, line 110: Can you add comments to this function essentially saying the object needs to have a valid `KlassID` in order to be considered sane. Maybe also mention the `openjdk/src/hotspot/share/oops/klass.hpp` file as well.

qinsoon: Sounds good. Done.

wks (approved):
LGTM

PR #196, opened 2023-02-07
Remove all #[inline] attributes

k-sareen:
This commit adds a section on making builds with profile-guided optimization to the README and also adds a script that makes a PGO build to the `.github/scripts` directory.


openjdk/CompileThirdPartyHeap.gmk
```diff
@@ -10,6 +10,9 @@ MMTK_CPP_ROOT = $(THIRD_PARTY_HEAP)
 OPENJDK_VERSION=`cd $(MMTK_RUST_ROOT) ; cargo read-manifest --manifest-path=Cargo.toml | python3 -c 'import json,sys; print(json.load(sys.stdin)["metadata"]["openjdk"]["openjdk_version"])'`
 OPENJDK_LOCAL_VERSION=`git rev-parse HEAD`
 
+# Get the current host triple. This is used for the PGO build
```
qinsoon, line 13: The change looks fine. But this seems to be a general change. How is this related with PGO?

README.md
```diff
@@ -126,6 +126,69 @@ $ make CONF=linux-x86_64-normal-server-release THIRD_PARTY_HEAP=$PWD/../mmtk-ope
...
+builds](https://doc.rust-lang.org/rustc/profile-guided-optimization.html) by
+directly hooking into the LLVM profiling infrastructure. In order to have the
+correct LLVM tools version, you should install the relevant `llvm-tools-preview`
+component using `rustup`:
+
+```console
+$ rustup component add llvm-tools-preview
+```
+
+We use `fop` as it is a relatively small benchmark but also exercises the GC.
```
qinsoon, line 142: This is specific to DaCapo, and specific to our approach. Ideally, you should run profiling with the exact workload you want to benchmark. Using `fop` for DaCapo is just our workaround to avoid running profiling for the whole DaCapo suite. It is okay to use `fop` as an example for the rest of the text. But probably make it clear here that for the best result, you should run the workload that you try to benchmark and use the exact MMTk plan you try to benchmark with.

README.md
```diff
@@ -126,6 +126,69 @@ $ make CONF=linux-x86_64-normal-server-release THIRD_PARTY_HEAP=$PWD/../mmtk-ope
...
+$ RUSTFLAGS="-Cprofile-use=/tmp/$USER/pgo-data/merged.profdata -Cllvm-args=-pgo-warn-missing-function" make CONF=linux-x86_64-normal-server-release THIRD_PARTY_HEAP=$PWD/../mmtk-openjdk/openjdk images
+```
+
+We now have an OpenJDK build under
+`./build/linux-x86_64-normal-server-release/images/jdk` with MMTk that has been
+optimized using PGO.
+
+For ease of use, we have provided a script which does the above in
+`.github/scripts/pgo-build.sh`. Note that you may have to change the location
+of `llvm-profdata`.
```
qinsoon, line 190: You can just say the script is an example, and you would need to change it to your need. There could be quite a few things that are varying to different users, like the Rust version, the location of the profile, the host triplet, the location of the Rust toolchain. It would be better not to give users an impression that they can just take the script, and run it.

qinsoon (approved):
LGTM. Just some minor issues.

openjdk/CompileThirdPartyHeap.gmk
k-sareen: I'm following what is mentioned in the Cargo PGO build guide: https://doc.rust-lang.org/rustc/profile-guided-optimization.html#a-complete-cargo-workflow. Specifically:
> We pass the --target flag to Cargo, which prevents the RUSTFLAGS arguments to be passed to Cargo build scripts. We don't want the build scripts to generate a bunch of .profraw files.

qinsoon: I see. Probably add the link in the comment so we won't get confused in the future (or attempt to remove that).

README.md
k-sareen: But for a general JVM build, we still have to profile against something. `fop` was just chosen as a general representative benchmark and `GenImmix` is one of our best-performing collectors. I would argue that if you tune too much to a particular benchmark, that's essentially giving MMTk an unfair edge as G1 and other GCs don't do PGO (or at least to the best of my knowledge).

README.md
k-sareen: The host-triplet should be automatically selected from `cargo -vV`. But yes, location of Rust toolchain etc. can be variable. I'll mention this is an example that they can further adapt for their uses.

README.md
qinsoon: You can keep most of the text here about `fop` and `GenImmix`. You just need to make it clear that this is just an example. As a user, you should profile with the plan you are using, and with the workload that you care about.

qinsoon:
@k-sareen Can you address the comments when you have time? Then we can get this merged. The other related PRs in MMTk core have been merged, and we would need this to be merged as well.

k-sareen:
Yes I will, sorry. Forgot I had to update the README.

k-sareen:
Done @qinsoon

PR #205, opened 2023-03-21
Add C2 slowpath allocation barrier

wenyuzhao:
[mmtk-core PR](https://github.com/mmtk/mmtk-core/pull/785)

Add the missing C2 slow-path allocation barrier.

The C2 slow-path allocation code can do deoptimization after the allocation and before returning to C2 compiled code. The deoptimization itself contains a safepoint.

For generational plans, if a GC happens at this safepoint, the allocated object will be promoted, and all the subsequent field initialization should be recorded. The solution is to unconditionally treat this object as being written if the object is not in nursery, and record this object in the modbuf. This is also the solution for G1: https://github.com/mmtk/openjdk/blob/e1e420b3c4fd627218e45ec977972561c559acec/src/hotspot/share/gc/shared/cardTableBarrierSet.cpp#L136-L160


openjdk/barriers/mmtkObjectBarrier.cpp
```diff
@@ -2,6 +2,21 @@
...
 
+void MMTkObjectBarrierSetRuntime::object_probable_write(oop new_obj) const {
+#if MMTK_ENABLE_BARRIER_FASTPATH
+  intptr_t addr = (intptr_t) (void*) new_obj;
+  uint8_t* meta_addr = (uint8_t*) (SIDE_METADATA_BASE_ADDRESS + (addr >> 6));
+  intptr_t shift = (addr >> 3) & 0b111;
+  uint8_t byte_val = *meta_addr;
+  if (((byte_val >> shift) & 1) == 1) {
+    // only promoted objects will reach here
+    mmtk_object_probable_write((MMTk_Mutator) &Thread::current()->third_party_heap_mutator, (void*) new_obj);
```
qinsoon, line 13: The method includes the fast path, and it will check the log bit again. Performance-wise it may not be a concern, but it can cause confusion. This also is different from the following normal write barrier code which calls different methods. If this is intended, please add some comments to explain the reason behind the decision.

wenyuzhao: Updated comments to explain the fast-path unlog bit check.

qinsoon (approved):

PR #206, opened 2023-03-22
Add new correctness CI

k-sareen:
This commit adds a new correctness CI which parallelizes builds and tests. It uses running-ng, the latest DaCapo snapshot, and reuses build artifacts across runs. All running-ng outputs are uploaded as artifacts at the end of the workflow.

### DIFF
--- a/.github/configs/base.yml
+++ b/.github/configs/base.yml
@@ -1,0 +1,86 @@
+includes:
+  - "$RUNNING_NG_PACKAGE_DATA/base/runbms.yml"
+
+suites:
+  dacapochopin-04132797-ci:
+    type: DaCapo
+    release: evaluation
+    path: "DACAPO_PATH/dacapo-evaluation-git-04132797.jar"
+    minheap: mmtk-openjdk-11-MarkCompact
+    minheap_values:
+      mmtk-openjdk-11-MarkCompact:
+        avrora: 8
+        batik: 426
+        biojava: 197
+        cassandra: 117
+        eclipse: 439
+        fop: 24
+        graphchi: 195
+        h2: 1122
+        h2o: 136
+        jme: 236
+        jython: 48
+        kafka: 243
+        luindex: 25
+        lusearch: 36
+        pmd: 291
+        spring: 110
+        sunflow: 37
+        tradebeans: .inf
+        tradesoap: .inf
+        tomcat: 55
+        xalan: 19
+        zxing: 427
+      nominal-heap-sizes:
+        avrora: 7
+        batik: 192
+        biojava: 97
+        cassandra: 74
+        eclipse: 346
+        fop: 17
+        graphchi: 179
+        h2: 506
+        h2o: 102
+        jme: 29
+        jython: 29
+        kafka: 207
+        luindex: 44
+        lusearch: 19
+        pmd: 114
+        spring: 47
+        sunflow: 25
+        tradebeans: 101
+        tradesoap: 89
+        tomcat: 17
+        xalan: 9
+        zxing: 66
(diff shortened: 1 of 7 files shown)


k-sareen:
I'm just checking the build works and then I'll cancel the CI so that we can cache the DaCapo benchmarks once.

k-sareen:
Currently it's a 15 min running-ng timeout, which I thought would have been fine, but we're almost certainly hitting it for some of the benchmarks. I don't know what's the best way to improve developer turnaround time and correctness. I think maybe we have a regular CI (using `cron`) that does full correctness and for PRs we have a reduced set of benchmarks. What do you think @caizixian @qinsoon @wks?

wks:
> Currently it's a 15 min running-ng timeout, which I thought would have been fine, but we're almost certainly hitting it for some of the benchmarks. I don't know what's the best way to improve developer turnaround time and correctness. I think maybe we have a regular CI (using `cron`) that does full correctness and for PRs we have a reduced set of benchmarks. What do you think @caizixian @qinsoon @wks?

If we introduce some kind of "watch dog" mechanism in DaCapo that "feeds the dog" between iterations, we can kill an unresponsive process at the precision of iteration instead of invocation.  If we run 5 iterations per invocation, we can reduce the timeout from 15 min to 3 min.

k-sareen:
The default is running 3 iterations x 3 invocations. Perhaps we can use the `--skip-timeout` functionality in running-ng to fail early if 1 benchmark timesout?

qinsoon:
> Currently it's a 15 min running-ng timeout

Is it the timeout for a single benchmark? That means, for one plan for 3 invocations, the worst time is 45mins. I think Github runners have a hard timeout limit of 6h. That means at maximum we can test 8 plans.

> The default is running 3 iterations x 3 invocations. Perhaps we can use the --skip-timeout functionality in running-ng to fail early if 1 benchmark times out?

It seems unnecessary to run 3 invocations. I remember I discussed this with Steve long time ago, and his response was there was no point to run a test multiple times in the CI just to catch random bugs. Running 3 iterations also may not be worth the time. If we just run 1x1, the runtime would be 1/9 of what it is now. I don't mind running more invocations. But if we would need to make the turnaround time shorter, this might be the most effective way to go.

k-sareen:
> It seems unnecessary to run 3 invocations. I remember I discussed this with Steve long time ago, and his response was there was no point to run a test multiple times in the CI just to catch random bugs. Running 3 iterations also may not be worth the time. If we just run 1x1, the runtime would be 1/9 of what it is now. I don't mind running more invocations. But if we would need to make the turnaround time shorter, this might be the most effective way to go.

I think it's useful to do one of more iterations and more invocations (preferably both). Gives us more confidence in the correctness.

k-sareen:
Also ***super important note*** is that the GitHub Cache expires if it hasn't been used in 7 days. The DaCapo benchmarks have been cached by forcing a solo run right above. But if the cache ever expires, we'll suddenly have 40 jobs all trying to set the cache. Either we should run a regular `cron` CI that will use the cache so that it will never expire or we add a new single job (i.e. not parallelized via matrix) just before the test matrix so that the DaCapo cache can be set only once if has expired.

k-sareen:
> Another point is that these tests are forcing C2 and upfront compilation -- maybe we don't want to do this as we want to catch regressions in the interpreter and C1 compiler as well. Maybe we just don't provide any compiler flags? This could potentially be faster as well.

It certainly is faster to not force C2, but then we miss out on C2-only bugs like for `luindex`.

caizixian (approved):
LTGM. I think we should merge it and fine-tune it later.

PR #208, opened 2023-03-27
Introduce DaCapo cache job dependency

k-sareen:
This commit also allows the old CI runs to be canceled if the PR is updated.


k-sareen:
@qinsoon @caizixian Any idea why the test-assertions job isn't running concurrently?

qinsoon:
It got cancelled. I am not sure what it means.
```console
Canceling since a higher priority waiting request for 'Dacapo Tests-cache-dacapo-first' exists
```

k-sareen:
Yeah I think the concurrency stuff is interfering. Also, why is it a separate file? I think it can just be integrated into the same `ci.yml` file.

qinsoon:
Possibly they are in the same concurrency group, and one job caused the other being cancelled.

k-sareen:
Ohh I think it's interfering with the concurrency in the new CI. Wait let me see if I can fix i

k-sareen:
Alright cool. Fixed @qinsoon. I note that the old CI is using DaCapo 2006 (I thought it was using DaCapo Bach) and is using ubuntu-18.04 which is EOL soon.

qinsoon:
> Alright cool. Fixed @qinsoon. I note that the old CI is using DaCapo 2006 (I thought it was using DaCapo Bach) and is using ubuntu-18.04 which is EOL soon. We should move them to Ubuntu 22.04 at least.

We are running these tests with 22.04 in mmtk-core. Moving to 22.04 should be trivial.

k-sareen:
I'll make a separate PR for reducing scope of tests after this PR has merged. We also need to have a separate regular (using [GitHub Actions `cron`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule)) correctness CI that is ran with our standard performance options as discussed. I don't think I'll have time to figure out how to present the delta as we probably want the delta to be for per GC per benchmark instead of the entire benchmark.

qinsoon:
> I'll make a separate PR for reducing scope of tests after this PR has merged. We also need to have a separate regular (using [GitHub Actions `cron`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule)) correctness CI that is ran with our standard performance options as discussed. I don't think I'll have time to figure out how to present the delta as we probably want the delta to be for per GC per benchmark instead of the entire benchmark.

Honestly I don't like cron jobs with Github. It is hard to track regression as we need to keep an eye on the results. I generally prefer running the tests with PR, or with commits. We discussed that we could just run 1 invocation. We probably can run both configuration with each PR with the same amount of time as it is now (3 invocations).

k-sareen:
> Honestly I don't like cron jobs with Github. It is hard to track regression as we need to keep an eye on the results. I generally prefer running the tests with PR, or with commits. We discussed that we could just run 1 invocation. We probably can run both configuration with each PR with the same amount of time as it is now (3 invocations).

I think GitHub emails you (though I don't know how it works with an organization) if your `cron` CI fails. Yeah I guess since we will only do 1 invocation for the correctness CI, it could be fast enough to run with each PR.

k-sareen:
Something like a stress test can't be run for every PR though. It only makes sense to do it continuously in the background. Maybe the correctness CI using the performance CI options can be done for every commit to master (as opposed to for every PR). There is no risk of multiple workers downloading the DaCapo benchmarks (after merging this PR), so even if the cache expires only one worker has to fetch it.

k-sareen:
Can we merge this @qinsoon? I'll be able to do the quick refactor to reduce invocations and iterations then

PR #210, opened 2023-03-27
Reduce invocations and iterations for correctness CI

k-sareen:


k-sareen:
This PR doesn't add the other correctness CI (i.e. the one which uses the same parameters as our performance runs). I can add it if we want.

caizixian (approved):

PR #211, opened 2023-03-28
Check results for new CI, allow some benchmarks to fail

qinsoon:
This PR adds a yaml file that records our expected results for DaCapo tests. It also adds a script to check the results for each CI run. We should generally expect a CI run to succeed (which means the results are expected) once this PR is merged. 

However, as we may have random failures that we haven't noticed, it is possible a CI run still fails. In such a case, we should update the `ci-expected-results.yml`, create an Github issue, and update https://github.com/mmtk/mmtk-openjdk/issues/246.

PR #212, opened 2023-03-29
Increase running timeout and clean up DaCapo cache job

k-sareen:


k-sareen:
Why is it caching the zip again...?

k-sareen:
Sigh: https://github.com/orgs/community/discussions/27059#discussioncomment-3254477
This is incredibly frustrating

caizixian (approved):

k-sareen:
Don't merge this yet. I need to look into which benchmarks are still failing due to a timeout and increase the timeout accordingly.

caizixian:
OK please coordinate with @qinsoon because he is compiling a yaml file for recording whether a test is meant to pass or fail.

@qinsoon I guess for the tests that fail, we can also include notes that say timeout, oom, etc.

qinsoon:
> @qinsoon I guess for the tests that fail, we can also include notes that say timeout, oom, etc.

Yeah, I just added that: https://github.com/mmtk/mmtk-openjdk/pull/211/commits/899a97ee2da8d112297bb6908f327b91cad4e34b. I am just checking the running script output such as `0abcde.g` with the yaml, and the script does not look into the actual log file to see if it is timeout or oom. The note just provides some extra information for the developers. Like when a test is expected to pass but failed, the script will print the note so it is easier for us to check and understand the results.

I am not sure if this thing is useful or not. I am still testing it.

k-sareen:
So these are the benchmarks that fail due to timeout:
`fastdebug`:
 - `biojava`: `SemiSpace`, `MarkSweep`, `MarkCompact`
 - `cassandra`: `MarkSweep`, `MarkCompact`
 - `eclipse`: `MarkSweep`, `MarkCompact`
 - `graphchi`: `MarkCompact`
 - `h2`: `Immix`, `MarkSweep`, `MarkCompact`
 - `h2o`: `MarkCompact`
 - `jython`: `MarkCompact`
 - `lusearch`: `MarkSweep`, `MarkCompact`
 - `pmd`: `MarkCompact`
 - `sunflow`: `MarkSweep`, `MarkCompact`
 - `tomcat`: `MarkCompact`
 - `xalan`: `MarkCompact`

qinsoon:
Do they timeout after 720s? `fastdebug` probably needs longer timeout. It uses a debug build of MMTk, and the GC is pretty slow.

k-sareen:
Yeah I've just increased the timeout to 20 min. But then some benchmarks may not give results in a timely fashion

k-sareen:
Even 20 min isn't enough. Sigh

qinsoon:
> @qinsoon What do we do? Even a 45 min timeout is not good enough for some GCs on some benchmarks. Maybe just reset the timer to 20 min and accept that these GCs can't run in the timeout?
> 
> EDIT: Or actually some extra benchmarks did finish in 45 min, it's just not all. I think we can keep the timeout at 45 min. But benchmarks like `h2` and `biojava` don't complete in time.
> 
> I can try increasing to 1 hour but then results for some benchmarks may not be as timely.

I don't mind if we keep increasing the timeout until we reach the 6h maximum timeout for a job for Github hosted runner. If we still cannot run some benchmarks in 6 hours, we may need to think about other solutions.

There are a few things we could do:
1. `fastdebug` is expected to be slow.  It uses unoptimized code, and it includes all the debug assertions. Especially when we run a slow GC plan, it would be worse. Expecting them to finish withing the same timeout as our release build is not reasonable. We can use a different configuration for `fastdebug`, and do things like reducing the number of iterations for `fastdebug`.
2. Some plans like mark compact seem to timeout for every benchmark, based on the list you posted above. We may look into the code to see if there is any assertion code that is too slow for the plan that we should make it as part of `extreme_assertions`.

k-sareen:
It's already 1 iteration 1 invocation. MarkCompact is just a really slow plan and 45 min timeout does seem to help as some benchmarks do finish in it.

k-sareen:
More benchmarks pass now. I'll check which ones are timeouts later

k-sareen:
Seems like `MarkSweep` is still failing for some benchmarks. I'll increase the timeout.

k-sareen:
@qinsoon Ah I see what's happening. The issue for cassandra for MarkSweep for example is the database load fails due to a timeout or something and so the benchmark is stuck at 95% complete. More benchmarks pass now (notably h2 which was one of the slowest) and I think increasing the timeout will not change what benchmarks pass.

The remaining benchmarks are failing due to OOMs, crashes, etc. I'll be merging this PR after lunch then.

PR #216, opened 2023-04-19
Fix invalid register values in `arraycopy_epilogue` barrier

wenyuzhao:
For the `arraycopy_epilogue` barrier, the `count` or `dst` register values may get overwritten after the array copy, and `arraycopy_epilogue` can receive invalid addresses.

This PR adds an `arraycopy_prologue` barrier, saves the register values in this prologue barrier, and restores them in `arraycopy_epilogue`.

This should fix generational plan crashes on some DaCapo benchmarks.

### DIFF
--- a/openjdk/barriers/mmtkObjectBarrier.cpp
+++ b/openjdk/barriers/mmtkObjectBarrier.cpp
@@ -82,9 +82,41 @@ void MMTkObjectBarrierSetAssembler::object_reference_write_post(MacroAssembler* 
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


wenyuzhao:
From the CI results, look like all the generational plans, including StickyImmix, passed all the DaCapo runs.

k-sareen:
Yeah it's safe to say from CI results that the bug for generational collectors is fixed. I'll check with our performance parameters + will check what OpenJDK GCs do

k-sareen (approved):
LGTM. Thanks!

PR #218, opened 2023-05-17
Update to MMTk core PR #817

qinsoon:
Update for https://github.com/mmtk/mmtk-core/pull/817.

### DIFF
--- a/mmtk/src/active_plan.rs
+++ b/mmtk/src/active_plan.rs
@@ -1,3 +1,4 @@
+use crate::abi::JavaThreadIteratorWithHandle;
 use crate::OpenJDK;
 use crate::SINGLETON;
 use crate::UPCALLS;
@@ -5,19 +6,25 @@ use mmtk::util::opaque_pointer::*;
 use mmtk::vm::ActivePlan;
 use mmtk::Mutator;
 use mmtk::Plan;
-use std::sync::{Mutex, MutexGuard};
+use std::marker::PhantomData;
+use std::mem::MaybeUninit;
 
 struct OpenJDKMutatorIterator<'a> {
-    _guard: MutexGuard<'a, ()>,
+    handle: MaybeUninit<JavaThreadIteratorWithHandle>,
+    _p: PhantomData<&'a ()>,
 }
 
 impl<'a> OpenJDKMutatorIterator<'a> {
-    fn new(guard: MutexGuard<'a, ()>) -> Self {
-        // Reset mutator iterator
+    fn new() -> Self {
+        let mut iter = Self {
+            handle: MaybeUninit::uninit(),
+            _p: PhantomData,
+        };
+        // Create JavaThreadIteratorWithHandle
         unsafe {
-            ((*UPCALLS).reset_mutator_iterator)();
+            ((*UPCALLS).new_java_thread_iterator)(iter.handle.as_mut_ptr());
         }
-        Self { _guard: guard }
+        iter
     }
 }
 
@@ -25,7 +32,7 @@ impl<'a> Iterator for OpenJDKMutatorIterator<'a> {
     type Item = &'a mut Mutator<OpenJDK>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        let next = unsafe { ((*UPCALLS).get_next_mutator)() };
+        let next = unsafe { ((*UPCALLS).java_thread_iterator_next)(self.handle.as_mut_ptr()) };
         if next.is_null() {
             None
         } else {
@@ -53,15 +60,10 @@ impl ActivePlan<OpenJDK> for VMActivePlan {
     }
 
     fn mutators<'a>() -> Box<dyn Iterator<Item = &'a mut Mutator<OpenJDK>> + 'a> {
-        let guard = MUTATOR_ITERATOR_LOCK.lock().unwrap();
-        Box::new(OpenJDKMutatorIterator::new(guard))
+        Box::new(OpenJDKMutatorIterator::new())
     }
 
     fn number_of_mutators() -> usize {
         unsafe { ((*UPCALLS).number_of_mutators)() }
     }
-}
-
-lazy_static! {
-    pub static ref MUTATOR_ITERATOR_LOCK: Mutex<()> = Mutex::new(());
 }
--- a/mmtk/src/active_plan.rs
+++ b/mmtk/src/active_plan.rs
@@ -5,7 +5,34 @@ use mmtk::util::opaque_pointer::*;
 use mmtk::vm::ActivePlan;
 use mmtk::Mutator;
 use mmtk::Plan;
-use std::sync::Mutex;
+use std::sync::{Mutex, MutexGuard};
+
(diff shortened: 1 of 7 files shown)


mmtk/src/active_plan.rs
```diff
@@ -5,7 +5,34 @@ use mmtk::util::opaque_pointer::*;
 use mmtk::vm::ActivePlan;
 use mmtk::Mutator;
 use mmtk::Plan;
-use std::sync::Mutex;
+use std::sync::{Mutex, MutexGuard};
+
+struct OpenJDKMutatorIterator<'a> {
+    _guard: MutexGuard<'a, ()>,
```
wks, line 11: Guarding the global iterator instance (`static MaybeUninit<JavaThreadIteratorWithHandle> jtiwh;`) in C++ is one way to adapt to the new API.  It is as correct as the previous code.

But given that OpenJDK provides a proper iterator type `JavaThreadIteratorWithHandle`, I think a more straightforward way to implement `OpenJDKMutatorIterator` is wrapping an instance of `JavaThreadIteratorWithHandle` (or a smart pointer to it) right into the `OpenJDKMutatorIterator` itself.

mmtk/src/active_plan.rs
```diff
@@ -5,7 +5,34 @@ use mmtk::util::opaque_pointer::*;
...
+        }
+        Self { _guard: guard }
+    }
+}
+
+impl<'a> Iterator for OpenJDKMutatorIterator<'a> {
+    type Item = &'a mut Mutator<OpenJDK>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let next = unsafe { ((*UPCALLS).get_next_mutator)() };
```
wks, line 28: This line is the same.  If we encapsulate the `JavaThreadIteratorWithHandle` instance in the `OpenJDKMutatorIterator` itself, we don't need to access the global iterator instance from here.

mmtk/src/active_plan.rs
```diff
@@ -5,7 +5,34 @@ use mmtk::util::opaque_pointer::*;
...
+
+struct OpenJDKMutatorIterator<'a> {
+    _guard: MutexGuard<'a, ()>,
+}
+
+impl<'a> OpenJDKMutatorIterator<'a> {
+    fn new(guard: MutexGuard<'a, ()>) -> Self {
+        // Reset mutator iterator
+        unsafe {
+            ((*UPCALLS).reset_mutator_iterator)();
```
wks, line 18: This line shows that an `OpenJDKMutatorIterator` instance is not really an independent instance.  All `OpenJDKMutatorIterator` instances are using the same global instance.

wks (changes requested):
Obviously this PR is a quick hack to adapt to the new API.  Although we can make it more idiomatic in a subsequent PR, I prefer we do it now, because implementing the API in an idiomatic way can help us check if this API is properly designed.  If we find it difficult to implement the `OpenJDKMutatorIterator` by wrapping a `JavaThreadIteratorWithHandle` (for example, due to lifetime issues), we may need to adjust the signature of the `ActivePlan::mutators()` method.

mmtk/src/active_plan.rs
qinsoon: Done. I removed the global iterator, and moved that to each iterator. `JavaThreadIteratorWithHandle` is a `StackObj`, thus cannot be allocated in the heap. I create an uninitialized instance in Rust, and pass that to C++ to let it initialize the instance.

openjdk/mmtkUpcalls.cpp
```diff
@@ -190,27 +190,18 @@ struct MaybeUninit {
   char _data[sizeof(T)];
 };
 
-static MaybeUninit<JavaThreadIteratorWithHandle> jtiwh;
-static bool mutator_iteration_start = true;
+static void mmtk_new_java_thread_iterator(void* iter) {
+  *(JavaThreadIteratorWithHandle*)iter = JavaThreadIteratorWithHandle();
```
wks, line 194: This step moves the constructed `JavaThreadIteratorWithHandle` instance to another location.

However, part of the race-prevention algorithm in OpenJDK includes letting each `Thread` hold a pointer to the `SafeThreadsListPtr` instance inside `JavaThreadIteratorWithHandle`.  If the `SafeThreadsListPtr` is moved, the field in `Thread` becomes a dangling pointer.

See

-   https://github.com/mmtk/openjdk/blob/e1e420b3c4fd627218e45ec977972561c559acec/src/hotspot/share/runtime/threadSMR.cpp#LL381C13-L381C13
-   https://github.com/mmtk/openjdk/blob/e1e420b3c4fd627218e45ec977972561c559acec/src/hotspot/share/runtime/threadSMR.cpp#L416

mmtk/src/lib.rs
```diff
@@ -95,6 +93,9 @@ pub struct OpenJDK_Upcalls {
     pub schedule_finalizer: extern "C" fn(),
     pub prepare_for_roots_re_scanning: extern "C" fn(),
     pub enqueue_references: extern "C" fn(objects: *const ObjectReference, len: usize),
+    pub new_java_thread_iterator: extern "C" fn(iter: *mut abi::JavaThreadIteratorWithHandle),
+    pub java_thread_iterator_next:
+        extern "C" fn(iter: *mut abi::JavaThreadIteratorWithHandle) -> *mut Mutator<OpenJDK>,
```
wks, line 98: There is one missing upcall: the destructor.  `JavaThreadIteratorWithHandle` has a destructor which should be called when the iterator is dropped.

wks (changes requested):
The obvious problem is that the Rust code now knows the layout of the C++ struct `JavaThreadIteratorWithHandle`.  This makes the code much less maintainable, especially if the layout changes in unspecified ways at different debug levels (and the Rust code must match the C++-side debug level).

From the code, I found that the `Thread` class has pointers back to some internal structures of `JavaThreadIteratorWithHandle` (see comments).  This prevents us from instantiating `JavaThreadIteratorWithHandle` on the stack and then copying it to the Rust `Iterator`.  I also tried working around this by using the [placement new]() syntax to initialise `JavaThreadIteratorWithHandle` in a pre-allocated buffer, but it failed too.  It looks like `StackObj` also prevented placement new by defining private `new` methods.

Perhaps the safest way is still switching to internal iteration.

wks:
Maybe the easiest solution is just collecting the mutator pointers in a `Vec` and embed the `Vec` into the `Iterator<Mutator>`.

The following is the typical use case in mmtk-core:

```rust
        for mutator in <C::VM as VMBinding>::VMActivePlan::mutators() {
            mmtk.scheduler.work_buckets[WorkBucketStage::Prepare]
                .add(PrepareMutator::<C::VM>::new(mutator));
        }
```

It creates one work packet for each mutator.  This means we assume the mutator pointer to be valid during the GC anyway, even after the iterator returned by `mutators()` returned.  Under such assumption, it should be safe to just collect the mutator pointers.

qinsoon:
Closed, in favor of https://github.com/mmtk/mmtk-openjdk/pull/220.

PR #220, opened 2023-05-22
Update MMTk core PR #817 (alternative approach)

wks:
This is an alternative approach for https://github.com/mmtk/mmtk-openjdk/pull/218 that addresses the upstream API change in https://github.com/mmtk/mmtk-core/pull/817.  This PR simply saves the mutator pointers into a `VecDeque` and embeds it into the iterator.


qinsoon (approved):
LGTM

PR #223, opened 2023-06-24
Update README.md and default local mmtk location

shamouda:


k-sareen:
@shamouda Thanks for the PR! I think we need to fix the CI scripts before we can merge this. Some of the CI scripts expect the previous location. I'll make another PR for it.

README.md
```diff
@@ -121,10 +121,10 @@ The output jdk is at `./build/linux-x86_64-normal-server-$DEBUG_LEVEL/jdk`.
 **Note:** The above `make` command will build what is known as the [`default` target or "exploded image"](https://github.com/openjdk/jdk11u/blob/master/doc/building.md#Running-make). This build is exclusively meant for developers who want quick and incremental builds to test changes. If you are planning on evaluating your build (be it performance, minimum heap, etc.), then it is *highly advised* to use the `images` target. The `default` target is the (roughly) minimal set of outputs required to run the built JDK and is not guaranteed to run all benchmarks. It may have bloated minimum heap values as well. The `images` target can be built like so:
 
 ```console
-$ make CONF=linux-x86_64-normal-server-release THIRD_PARTY_HEAP=$PWD/../mmtk-openjdk/openjdk images
```
k-sareen, line 124: Could you add a line saying "For performance evaluation, we expect you are using the `release` build"? That was the original reason the `make` command was not using `$DEBUG_LEVEL` and directly using `release`. I agree that it should have been sign posted a bit better.

k-sareen: Or better yet, mention it in the note directly above the `make` command.

shamouda: Does the same applies to the "Profile-Guided Optimized Build" section? 
If that is the case, then I should probably either:
1. undo all the changes (i.e. restore the hardcoded -release values) but add a clarifying note, or
2. add "DEBUG_LEVEL=release" before the `make` commands to make it more explicit.

What do you think?

k-sareen: I think the first one is probably better. And yes, the PGO build is also generally used for evaluation only. The `images` (and PGO build) have extra steps involved in them which only make sense when you want to perform some kind of evaluation. In the `images` case, the `default` build (an unoptimized dev build) is used to bootstrap a production build and in the PGO case, the Rust compiler use execution profiles to inform inlining decisions.

k-sareen:
@qinsoon Which CI script is even using the hardcoded value? I don't see it being used by the old CI script here? Clearly it's passing the tests above.

qinsoon:
> @qinsoon Which CI script is even using the hardcoded value? I don't see it being used by the old CI script here? Clearly it's passing the tests above.

https://github.com/mmtk/mmtk-core/blob/40c977308cb24d8bdf46b82916a388c10fa956e7/.github/workflows/post-review-ci.yml#L77-L86

We can update the mmtk-core CI to parse and modify the toml semantically. In that case, it doesn't matter which value we put in the toml file. See https://github.com/mmtk/mmtk-core/issues/856

k-sareen:
Ah right it's in the mmtk-core CI script. I see. Yes we should parse it in a more robust way imo. I'll investigate how best to do it tonight. `jq` sounds like the best idea.

qinsoon:
> Ah right it's in the mmtk-core CI script. I see. Yes we should parse it in a more robust way imo. I'll investigate how best to do it tonight. `jq` sounds like the best idea.

What Kunshan posted is about reading data from toml. So we can use `cargo metadata` to parse toml as json, and use jq to read it. Here we need to write to toml.

k-sareen:
> What Kunshan posted is about reading data from toml. So we can use `cargo metadata` to parse toml as json, and use jq to read it. Here we need to write to toml.

Yes that makes sense. How about using a slightly non-standard tool like this: https://github.com/TomWright/dasel. I found it by doing a bit of quick googling.

wks:
I think when we modify, we should directly modify the TOML file, not the generated JSON from `cargo metadata`.  `cargo metadata` also aggregates implied information not included in the TOML file, including metadata of dependencies.

There are TOML libraries for Python, too.

Alternatively, there is a tool named `fq` (https://github.com/wader/fq) which can parse many formats including TOML.  But it is a bit unstable and not very well-documented.

k-sareen:
That's why I suggested the `dasel` tool. Seems fairly stable and works directly on TOML files.

PR #228, opened 2023-08-15
Add features for genimmix and stickyimmix

qinsoon:
We currently allow build with a fixed plan to work with the current performance CI. https://github.com/mmtk/mmtk-core/pull/901 attempts to add more plans to our performance CI. This PR adds some missing features for genimmix and stickyimmix. 

We do not need those plans once we move to the new `running-ng` scripts.


wks (approved):
LGTM.

PR #229, opened 2023-08-17
Pass hotspot command line flags to mmtk-core

wenyuzhao:
This PR makes mmtk-core respect to some hotspot command line flags. 

Currently, for mmtk options `threads` and `transparent_hugepage`, mmtk-openjdk will apply the values from hotspot flags (`ParallelGCThreads` and `UseTransparentHugePages` respectively) first. Specifying these options in `ThirdPartyHeapOptions` or env vars can still work and will override the values copied from hotspot flags.

### DIFF
--- a/mmtk/src/api.rs
+++ b/mmtk/src/api.rs
@@ -18,6 +18,7 @@ use mmtk::MutatorContext;
 use once_cell::sync;
 use std::cell::RefCell;
 use std::ffi::{CStr, CString};
+use std::fmt::Display;
 use std::sync::atomic::Ordering;
 
 // Supported barriers:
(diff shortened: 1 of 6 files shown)


mmtk/src/api.rs
```diff
@@ -290,6 +291,37 @@ pub extern "C" fn process(name: *const c_char, value: *const c_char) -> bool {
...
+pub extern "C" fn mmtk_set_hotspot_flag_uint(name: *const c_char, value: u32) {
+    set_hotspot_flag_impl(name, value)
+}
+
+/// Pass a hotspot boolean command line flag to mmtk
+#[no_mangle]
+pub extern "C" fn mmtk_set_hotspot_flag_bool(name: *const c_char, value: bool) {
+    set_hotspot_flag_impl(name, value)
+}
+
```
wks, line 324: In fact the `options` field in `MMTKBuilder` is mutable, and can be set in a strongly typed way.  And the code is quite simple.  It is pointless to parse options by name.

```suggestion
#[no_mangle]
pub extern "C" fn mmtk_builder_set_threads(value: usize) {
    let mut builder = BUILDER.lock().unwrap();
    builder.options.threads.set(value);
}

#[no_mangle]
pub extern "C" fn mmtk_builder_set_transparent_hugepages(value: bool) {
    let mut builder = BUILDER.lock().unwrap();
    builder.options.transparent_hugepages.set(value);
}

```

wenyuzhao: Thanks. Fixed. I didn't release there is a typed interface to do that.

wks (approved):
LGTM

mmtk/src/api.rs
wks: > ... I didn't release there is a typed interface to do that.

It was made possible since https://github.com/mmtk/mmtk-core/pull/625.  Before that, the `Options` struct was within an `UnsafeOptionsWrapper` which only allowed the string-based setters.  After that, the `Options` is mutable inside `MMTKBuilder`, and immutable in `MMTK`.  That allows it to be manipulated directly when still in `MMTKBuilder`.

Issue #231, opened 2023-08-18
Misaligned pointer dereference after moving to Rust 1.71.1


qinsoon:
OpenJDK supplies unaligned address as an edge. This becomes an error in debug build in Rust 1.71.1

```console
thread '<unnamed>' panicked at 'misaligned pointer dereference: address must be a multiple of 0x8 but is 0x7f569d60ed36', /home/runner/.cargo/git/checkouts/mmtk-core-89cdc7bf360cce7f/e88caa1/src/util/address.rs:220:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at 'panic in a function that cannot unwind', library/core/src/panicking.rs:126:5
stack backtrace:
   0:     0x7f56b2035861 - std::backtrace_rs::backtrace::libunwind::trace::h782cc21a5acaf6cb
                               at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x7f56b2035861 - std::backtrace_rs::backtrace::trace_unsynchronized::hc579eb24ab204515
                               at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x7f56b2035861 - std::sys_common::backtrace::_print_fmt::h7223525cfdbacda2
                               at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/sys_common/backtrace.rs:65:5
   3:     0x7f56b2035861 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hbd7d55b7108d2ab8
                               at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x7f56b205945f - core::fmt::rt::Argument::fmt::hb4f4a02b9bd9dd49
                               at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/fmt/rt.rs:138:9
   5:     0x7f56b205945f - core::fmt::write::h6d54cd7c9e155ec5
                               at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/fmt/mod.rs:1094:21
[...]

PR #232, opened 2023-08-21
Fix unaligned edge access

wks:
This allows the OpenJDK binding to load and store ObjectReferences from/to edges at unaligned addresses on x86 and x86_64.

Fixes: https://github.com/mmtk/mmtk-openjdk/issues/231

### DIFF
--- a/mmtk/src/api.rs
+++ b/mmtk/src/api.rs
@@ -343,7 +343,7 @@ pub extern "C" fn mmtk_object_reference_write_pre(
 ) {
     mutator
         .barrier()
-        .object_reference_write_pre(src, slot, target);
+        .object_reference_write_pre(src, slot.into(), target);
 }
 
 /// Full post barrier
@@ -356,7 +356,7 @@ pub extern "C" fn mmtk_object_reference_write_post(
 ) {
     mutator
         .barrier()
-        .object_reference_write_post(src, slot, target);
+        .object_reference_write_post(src, slot.into(), target);
 }
 
 /// Barrier slow-path call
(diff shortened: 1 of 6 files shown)


qinsoon (approved):
LGTM. Thanks.

wks:
https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2023-08-21-Mon-135824&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.stw&|10&iteration^1^4|20&1^invocation|30&1&mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&

I ran `lusearch`, 100 iterations, 2.5x min heap size, on `bobcat.moma`.  Build1 is master and build2 is this PR.

The difference in STW time is less than 1%.  There seems to be a very small amount of performance improvement, but is within the confidence interval.

image

wenyuzhao (approved):

wks:
The generated assembly for `ProcessEdgesWork::do_work` is identical between build1 and build2.  I think the difference is a result of noise or subtle differences in unrelated code.

```asm
; r14: self; rax: buffer
  109ca1:	49 8b 46 08          	mov    0x8(%r14),%rax      ; let buffer = self.edges.buf;
; rbx: index; r13: edge
  109ca5:	4c 8b 2c d8          	mov    (%rax,%rbx,8),%r13  ; let edge = buffer[index * 8];
  109ca9:	48 ff c3             	inc    %rbx                ; index += 1
; rdx: object
  109cac:	49 8b 55 00          	mov    0x0(%r13),%rdx      ; let object = edge.addr.as_ptr().read_unaligned();
  109cb0:	48 85 d2             	test   %rdx,%rdx           ; if object.is_null() {
  109cb3:	75 cb                	jne    109c80 <mmtk::scheduler::gc_work::<impl mmtk::scheduler::work::GCWork<<E as mmtk::scheduler::gc_work::ProcessEdgesWork>::VM> for E>::do_work+0x30>
```

PR #235, opened 2023-08-30
Compressed Oops Support

wenyuzhao:
This PR adds compressed oops support for mmtk-openjdk, and enables it by default.

# Implementation strategy and workarounds

## Heap layout and compression policy

This PR uses the `Edge` type to abstract over the compressed edge and uncompressed edge. Object field loads and stores for uncompressed edges work as before. Loads and stores for compressed edges will involve an additional compression and decompression step.

In general, this is the function to decode a 32-bit compressed pointer to its uncompressed form:

```rust
fn decode(compressed_oop: u32) -> u64 {
    BASE + ((compressed_oop as u64) << SHIFT)
}
```

OpenJDK has a few optimizations to reduce the add and shift operations in JIT-compiled code, this PR supports them all:

1. For heap <= 3G, if we set the heap range as `0x4000_0000..0x1_0000_0000`, it is possible to totally remove the add and the shift. The compressed and uncompressed forms are identical.
    * Set `BASE = 0` and `SHIFT = 0` for this case.
2. For heap <= 31G, if we set the heap range as `0x4000_0000..0x8_0000_0000`, it is possible to remove the add.
    * Set `BASE = 0` and `SHIFT = 3` for this case.
3. For heap > 31G, the add and shift operation is still necessary.

For cases (1) and (2), the jit compiled code will contain less or even no encoding/decoding instructions, and hence improve the mutator performance. However, in Rust code, we still do the add and shift unconditionally, even when `BASE` or `SHIFT` is set to zero.

## NULL pointer checking

Generally, `BASE` can be any address as long as the memory is not reserved by others. However, `BASE` must be smaller than `HEAP_START`, otherwise `HEAP_START` will be encoded as `0` and be treated as a null pointer.

Same as openjdk, we set `BASE` to `HEAP_START - 4096` to solve this issue.

## Type specialization

Since we only support one edge type per binding, providing two `OpenJDKEdge` in one `MMTK` instance is not possible. 

This PR solves the issue by specializing almost all the types in the binding, with a `const COMPRESSED: bool` generic type argument. It provides two `MMTK` singletons: `MMTK<OpenJDK<COMPRESSED = true>>` and `MMTK<OpenJDK<COMPRESSED = false>>`. `MMTK<OpenJDK<COMPRESSED = true>>` will have the `OpenJDKEdge<COMPRESSED = true>` edge that does the extra pointer compression/decompression.

The two MMTK singletons are wrapped in two lazy_static global variables. The binding will only initialize one of them depending on the OpenJDK command-line arguments. Initializing the wrong one that does not match the `UseCompressedOops` flag will trigger an assertion failure.

## Pointer tagging

When compressed oops is enabled, all the fields are guaranteed to be compressed oops. However, stack or other global root pointers may be still uncompressed. The GC needs to handle both compressed and uncompressed edges and be able to distinguish between them.

To support this, this PR treats all the root `OpenJDKEdge<COMPRESSED = true>`s as tagged pointers. If the 63-th bit is set, this indicates that this edge points to a 64-nit uncompressed oop, instead of a compressed oop. And the `OpenJDKEdge::{load, store}` methods will skip the encoding/decoding step.

For object field edges, the encoding is performed unconditionally without the pointer tag check.

When compressed oops is disabled, there is no pointer tag check as well.

## Embedded pointers

Some (or probably all) pointers embedded in code objects are also compressed. On x64, it is always compressed to a `u32` integer that sits in an unaligned memory location. This means we need to (1) treat them as compressed oops just like other roots. (2) still performs the unaligned stores and loads.

However, for other architectures, the compressed embedded pointers may not be encoded as a `u32` anymore.

## Compressed `Klass*` pointers

When `UseCompressedOops` is enabled, by default it also enables `UseCompressedClassPointers`. This will make the `Klass*` pointer in the object header compressed to a `u32` as well. This PR supports class pointer compression as well. 

However, class pointer compression is only supported and tested when the compressed oops is enabled. The two flags must be enabled or disabled together. Enabling only one of them is not tested, not supported, and will trigger a runtime assertion failure.

---

# Performance results

[SemiSpace](http://squirrel.anu.edu.au/plotty-public/wenyuz/v8/p/mm26Ra)
[Immix](http://squirrel.anu.edu.au/plotty-public/wenyuz/v8/p/wEDPv4)


wenyuzhao:
The PR is ready for review. Sometimes the CI job `DaCapo Tests / test` or `DaCapo Tests / test-assertions` can be cancelled for no reason... rerun it for a few more times can make it pass.

qinsoon:
> The PR is ready for review. Sometimes the CI job `DaCapo Tests / test` or `DaCapo Tests / test-assertions` can be cancelled for no reason... rerun it for a few more times can make it pass.

These are the runs that failed.
https://github.com/mmtk/mmtk-openjdk/actions/runs/6261214028/attempts/1
https://github.com/mmtk/mmtk-openjdk/actions/runs/6261214028/attempts/2

Error messages are:
```console
The hosted runner: GitHub Actions 3 lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.
```
```console
The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.
```

If this only happened in a certain period of time, it simply could be a Github issue.

qinsoon:
`test-assertions` failed again in https://github.com/mmtk/mmtk-openjdk/actions/runs/6293506467/job/17084111016 with the following error message. Github did not have reported performance downgrade when the job failed (https://www.githubstatus.com/). So it should not be a Github outage.
```console
The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.
```

There are many discussions about this error for Github runners. It usually means the resources consumed in the process exceeded the limits. It could be related with the code change in the PR (if it consumes more memory or more CPU). It could also be changes in the Github runner specs. The simplest thing to try is to create another PR without any code change, and see if we see this error.

mmtk/src/abi.rs
```diff
@@ -356,34 +399,58 @@ pub struct ArrayOopDesc(OopDesc);
...
+            *((self as *const Self as *const u8).add(Self::length_offset::<COMPRESSED>())
+                as *const i32)
+        }
     }
-    fn base(&self, ty: BasicType) -> Address {
-        let base_offset_in_bytes = Self::header_size(ty) * BYTES_IN_WORD;
-        Address::from_ptr(unsafe { (self as *const _ as *const u8).add(base_offset_in_bytes) })
+    fn base<const COMPRESSED: bool>(&self, ty: BasicType) -> Address {
+        let base_offset_in_bytes = Self::header_size::<COMPRESSED>(ty) * BYTES_IN_WORD;
+        Address::from_ptr(unsafe { (self as *const Self as *const u8).add(base_offset_in_bytes) })
```
wks, line 433: Ditto.  Given we are using `Address` anyway, it is pointless to cast one pointer to another just for doing address arithmetic.

```suggestion
        Address::from_ref(self) + base_offset_in_bytes
```

mmtk/src/abi.rs
```diff
@@ -356,34 +399,58 @@ pub struct ArrayOopDesc(OopDesc);
...
+        )
+    }
+
+    pub unsafe fn slice<const COMPRESSED: bool>(
+        &self,
+        ty: BasicType,
+    ) -> crate::OpenJDKEdgeRange<COMPRESSED> {
+        let base = self.base::<COMPRESSED>(ty);
+        let start = base;
+        let end = base + ((self.length::<COMPRESSED>() as usize) << if COMPRESSED { 2 } else { 3 });
```
wks, line 452: We can use the `LOG_BYTES_IN_EDGE` constant to avoid duplication.

```suggestion
        let lshift = OpenJDKEdge::<COMPRESSED>::LOG_BYTES_IN_EDGE;
        let end = base + ((self.length::<COMPRESSED>() as usize) << lshift);
```

mmtk/src/abi.rs
```diff
@@ -168,7 +172,7 @@ impl InstanceKlass {
     const VTABLE_START_OFFSET: usize = Self::HEADER_SIZE * BYTES_IN_WORD;
 
     fn start_of_vtable(&self) -> *const usize {
-        unsafe { (self as *const _ as *const u8).add(Self::VTABLE_START_OFFSET) as _ }
+        unsafe { (self as *const Self as *const u8).add(Self::VTABLE_START_OFFSET) as _ }
```
wks, line 175: Should be easier to use `Address` here, too.  The function `start_of_itable` below can be changed similarly, too.

```suggestion
        (Address::from_ref(self) + Self::VTABLE_START_OFFSET).to_ptr()
```

mmtk/src/object_model.rs
```diff
@@ -48,9 +48,9 @@ impl ObjectModel<OpenJDK> for VMObjectModel {
                 unsafe { (dst + i).store((src + i).load::<u8>()) };
             }
         }
-        let start = Self::ref_to_object_start(to);
+        let start = <Self as ObjectModel<OpenJDK<COMPRESSED>>>::ref_to_object_start(to);
```
wks, line 51: The `<Self as .....>` is necessary because the `VMObjectModel` type actually implements both `ObjectModel<OpenJDK<true>>` and `ObjectModel<OpenJDK<false>>`.

This can be made unnecessary by giving `struct VMObjectModel` a const parameter:

```rust
pub struct VMObjectModel<const COMPRESSED: bool> {}
```

and let it only implement the trait with the same const argument:

```rust
impl<const COMPRESSED: bool> ObjectModel<OpenJDK<COMPRESSED>> for VMObjectModel<COMPRESSED> {
```

Then `VMObjectModel<true>` only implements `ObjectModel<OpenJDK<true>>`, and `VMObjectModel<false>` only implements `ObjectModel<OpenJDK<false>>`.

But the current code `<Self as ObjectModel<OpenJDK<COMPRESSED>>>` is still OK given that it does not appear too often.

mmtk/src/object_model.rs
```diff
@@ -60,11 +60,15 @@ impl ObjectModel<OpenJDK> for VMObjectModel {
...
+        if COMPRESSED {
+            unsafe { Oop::from(object).size::<true>() }
+        } else {
+            unsafe { Oop::from(object).size::<false>() }
+        }
     }
 
     fn get_size_when_copied(object: ObjectReference) -> usize {
-        Self::get_current_size(object)
+        <Self as ObjectModel<OpenJDK<COMPRESSED>>>::get_current_size(object)
```
wks, line 71: Ditto.

mmtk/src/object_scanning.rs
```diff
@@ -128,69 +145,96 @@ impl OopIterate for InstanceRefKlass {
...
-            let array_klass = unsafe { oop.klass.cast::<ObjArrayKlass>() };
-            array_klass.oop_iterate(oop, closure);
+            let array_klass = unsafe { klass.cast::<ObjArrayKlass>() };
+            array_klass.oop_iterate::<COMPRESSED>(oop, closure);
         }
         KlassID::TypeArray => {
-            let array_klass = unsafe { oop.klass.cast::<TypeArrayKlass>() };
-            array_klass.oop_iterate(oop, closure);
+            //     let array_klass = unsafe { oop.klass::<COMPRESSED>().cast::<TypeArrayKlass>() };
+            //     array_klass.oop_iterate::<C, COMPRESSED>(oop, closure);
```
wks, line 210: Are these lines intentionally commented out, or they are just commented out for testing?

mmtk/src/abi.rs
```diff
@@ -356,34 +399,58 @@ pub struct ArrayOopDesc(OopDesc);
...
         } else {
             typesize_in_bytes / BYTES_IN_WORD
         }
     }
-    fn length(&self) -> i32 {
-        unsafe { *((self as *const _ as *const u8).add(Self::LENGTH_OFFSET) as *const i32) }
+    fn length<const COMPRESSED: bool>(&self) -> i32 {
+        unsafe {
+            *((self as *const Self as *const u8).add(Self::length_offset::<COMPRESSED>())
+                as *const i32)
```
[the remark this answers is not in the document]
wenyuzhao: Fixed

mmtk/src/abi.rs
wenyuzhao: Fixed

mmtk/src/abi.rs
wenyuzhao: Fixed

mmtk/src/abi.rs
wenyuzhao: Fixed

mmtk/src/edges.rs
```diff
@@ -1,103 +1,269 @@
...
+pub fn initialize_compressed_oops_base_and_shift() {
+    let heap_end = mmtk::memory_manager::last_heap_address().as_usize();
+    if heap_end <= (4usize << 30) {
+        BASE.store(Address::ZERO, Ordering::Relaxed);
+        SHIFT.store(0, Ordering::Relaxed);
+    } else if heap_end <= (32usize << 30) {
+        BASE.store(Address::ZERO, Ordering::Relaxed);
+        SHIFT.store(3, Ordering::Relaxed);
+    } else {
+        // set heap base as HEAP_START - 4096, to make sure null pointer value is not conflict with HEAP_START
```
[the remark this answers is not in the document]
wenyuzhao: Fixed

mmtk/src/edges.rs
```diff
@@ -1,103 +1,269 @@
...
 }
 
-impl Edge for OpenJDKEdge {
+/// For OpenJDK<COMPRESSED = false>, every edge is uncompressed.
+///
+/// For COMPRESSED = true,
+///     * if this is a field of an object, the edge is compressed
+///     * if this is a root pointer: The c++ part of the binding shoud pass all the root pointers to rust as tagged pointers.
+///       If the 63-th bit of the pointer is set to 1, the value referenced by the pointer is a 32-bit compressed integer.
+///       Otherwise, it is a uncompressed root pointer.
```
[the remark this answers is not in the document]
wenyuzhao: Fixed

mmtk/src/abi.rs
```diff
@@ -356,34 +399,58 @@ pub struct ArrayOopDesc(OopDesc);
...
-    // This provides an easy way to access the array data in Rust. However, the array data
-    // is Java types, so we have to map Java types to Rust types. The caller needs to guarantee:
-    // 1. <T> matches the actual Java type
-    // 2. <T> matches the argument, BasicType `ty`
-    pub unsafe fn data<T>(&self, ty: BasicType) -> &[T] {
-        slice::from_raw_parts(self.base(ty).to_ptr(), self.length() as _)
+    /// This provides an easy way to access the array data in Rust. However, the array data
+    /// is Java types, so we have to map Java types to Rust types. The caller needs to guarantee:
+    /// 1. <T> matches the actual Java type
+    /// 2. <T> matches the argument, BasicType `ty`
```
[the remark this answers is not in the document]
wenyuzhao: Fixed

mmtk/src/edges.rs
```diff
@@ -1,103 +1,269 @@
...
     }
 
     fn copy(src: &Self, tgt: &Self) {
-        MemorySlice::copy(&src.range, &tgt.range)
+        debug_assert_eq!(src.bytes(), tgt.bytes());
+        debug_assert_eq!(
+            src.bytes() & ((1 << LOG_BYTES_IN_INT) - 1),
+            0,
+            "bytes are not a multiple of 32-bit integers"
+        );
```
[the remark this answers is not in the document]
wenyuzhao: Fixed

mmtk/src/object_model.rs
```diff
@@ -60,11 +60,15 @@ impl ObjectModel<OpenJDK> for VMObjectModel {
     }
 
     fn get_current_size(object: ObjectReference) -> usize {
-        unsafe { Oop::from(object).size() }
+        if COMPRESSED {
+            unsafe { Oop::from(object).size::<true>() }
+        } else {
+            unsafe { Oop::from(object).size::<false>() }
+        }
```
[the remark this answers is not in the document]
wenyuzhao: Fixed

mmtk/src/object_model.rs
wenyuzhao: Fixed

mmtk/src/object_model.rs
wenyuzhao: Fixed

mmtk/src/object_scanning.rs
wenyuzhao: Removed. This is just a debugging left-over

mmtk/src/abi.rs
```diff
@@ -356,34 +399,58 @@ pub struct ArrayOopDesc(OopDesc);
 pub type ArrayOop = &'static ArrayOopDesc;
 
 impl ArrayOopDesc {
-    const LENGTH_OFFSET: usize = mem::size_of::<Self>();
+    fn length_offset<const COMPRESSED: bool>() -> usize {
+        if COMPRESSED {
+            mem::size_of::<usize>() + mem::size_of::<u32>()
```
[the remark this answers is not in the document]
wenyuzhao: Fixed. the length offset is now calculated as `memoffset::offset_of!(OopDesc, klass) + mem::size_of::<NarrowKlass>()`

openjdk/mmtkHeap.cpp
```diff
@@ -380,12 +387,16 @@ void MMTkHeap::print_tracing_info() const {
 // Registering and unregistering an nmethod (compiled code) with the heap.
 // Override with specific mechanism for each specialized heap type.
 class MMTkRegisterNMethodOopClosure: public OopClosure {
-  template <class T> void do_oop_work(T* p) {
+  template <class T> void do_oop_work(T* p, bool narrow) {
+    if (UseCompressedOops && !narrow) {
+      guarantee((uintptr_t(p) & (1ull << 63)) == 0, "test");
+      p = (T*) (uintptr_t(p) | (1ull << 63));
```
wks, line 393: I am being pedantic, but it is usually considered a bad practice to overwrite function arguments.  It is safer to introduce new variables.  It avoids bugs like this one: https://github.com/mmtk/mmtk-core/pull/864#discussion_r1294167769

wks (approved):
LGTM.

I also left a comment, but it may be too pedantic.

openjdk/mmtkHeap.cpp
wenyuzhao: Fixed.

PR #236, opened 2023-09-01
Update pgo-build script to use the pinned Rust toolchain

caizixian:


caizixian:
Actually, we might need to hold off this PR because PGO+LTO is broken on Rust 1.71.1

https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/PGO/near/388531250
https://github.com/rust-lang/rust/issues/115344

caizixian:
We should include PGO in the CI

.github/scripts/pgo-build.sh
```diff
@@ -1,16 +1,24 @@
 #!/bin/bash
 
+pushd ../mmtk-openjdk/mmtk
+cargo clean
```
qinsoon, line 4: Why do we need `cargo clean`? Is it to work around the compiler bug? If so, probably add a comment so we may delete them once we move to a newer version.

qinsoon (approved):
LGTM

PR #243, opened 2023-09-13
Overwrite threads and transparent hugepage options only if they're default values

### LINKED ISSUE #242: `MMTK_THREADS` is ignored after #229

wenyuzhao:
Fixes #242


wenyuzhao:
Close this for now as we have a better solution.

PR #244, opened 2023-09-13
Fix ignored env var options

wks:
This PR lets MMTk-side Options read environment variables for options after OpenJDK-side options are applied.  By doing this, environment variables like MMTK_THREADS are no longer ignored.

Fixes: https://github.com/mmtk/mmtk-openjdk/issues/242

This PR makes use of the change introduced in https://github.com/mmtk/mmtk-core/pull/955

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,36 @@ $ ./build/linux-x86_64-normal-server-$DEBUG_LEVEL/jdk/bin/java HelloWorld
 Hello World!
 ```
 
+### Run HelloWorld (with MMTk)
+
+Pass `-XX:+UseThirdPartyHeap` as java command line arguments to enable MMTk.
+
+```
+$ ./build/linux-x86_64-normal-server-$DEBUG_LEVEL/jdk/bin/java -XX:+UseThirdPartyHeap HelloWorld
+```
+
+If `DEBUG_LEVEL` is `release`, you should just see
+
+```
+Hello World!
+```
+
+If `DEBUG_LEVEL` has other values (such as `slowdebug`), you should see logs, too.
+
+```
+[2023-09-14T06:18:46Z INFO  mmtk::memory_manager] Initialized MMTk with GenImmix (DynamicHeapSize(6815744, 8377073664))
+[2023-09-14T06:18:47Z INFO  mmtk::util::heap::gc_trigger] [POLL] nursery: Triggering collection (1670/1664 pages)
+[2023-09-14T06:18:47Z INFO  mmtk::plan::generational::global] Nursery GC
+[2023-09-14T06:18:47Z INFO  mmtk::scheduler::gc_work] End of GC (304/1664 pages, took 19 ms)
+[2023-09-14T06:18:47Z INFO  mmtk::util::heap::gc_trigger] [POLL] nursery: Triggering collection (1680/1664 pages)
+[2023-09-14T06:18:47Z INFO  mmtk::plan::generational::global] Nursery GC
+[2023-09-14T06:18:47Z INFO  mmtk::scheduler::gc_work] End of GC (460/1664 pages, took 11 ms)
+[2023-09-14T06:18:47Z INFO  mmtk::util::heap::gc_trigger] [POLL] nursery: Triggering collection (1674/1664 pages)
+[2023-09-14T06:18:47Z INFO  mmtk::plan::generational::global] Nursery GC
+[2023-09-14T06:18:47Z INFO  mmtk::scheduler::gc_work] End of GC (614/1664 pages, took 11 ms)
+Hello World!
+```
+
 ### Run DaCapo Benchmarks with MMTk
 
 First, fetch DaCapo:
@@ -255,4 +285,25 @@ Using scaled threading model. 24 processors detected, 24 threads used to drive t
 ===== DaCapo 9.12-MR1 lusearch PASSED in 822 msec =====
 ```
 
-**Note:** Pass `-XX:+UseThirdPartyHeap` as java command line arguments to enable MMTk.
+### MMTk options
+
+MMTk has many options defined in https://github.com/mmtk/mmtk-core/blob/master/src/util/options.rs
+
+You can use environment variables started with `MMTK_` to set those options.  For example,
+`export MMTK_THREADS=1` will set the number of GC worker threads to one.  Follow the link above for
+more details.
+
+You can also set those options via command line arguments.
+
+-   `-XX:ParallelGCThreads=n` (where `n` is a number) sets the number of GC worker threads.
+    -   MMTk option: `Options::threads`
+-   `-XX:+UseTransparentHugePages` enables transparent huge pages.
+    -   MMTk option: `Options::transparent_hugepages`
+
+You can also set MMTk options using `-XX:ThirdPartyHeapOptions=options`, where `options` is
+`key=value` pairs separated by commas (`,`).  For example, `stress_factor=1000000,threads=1` will
+set `stress_factor` to 1000000, and `threads` to 1.
+
+Options set via command line arguments take prioritiy over environment variables starting with
+`MMTK_`.  If both the environment variable `MMTK_THREADS=1` and the command line argument
+`-XX:ParallelGCThreads=2` are give, the numberof GC worker threads will be 2.
--- a/README.md
+++ b/README.md
@@ -293,16 +293,19 @@ You can use environment variables started with `MMTK_` to set those options.  Fo
 `export MMTK_THREADS=1` will set the number of GC worker threads to one.  Follow the link above for
 more details.
 
-You can also set those options via command line arguments.
+You can also set those options via command line arguments: `-XX:ThirdPartyHeapOptions=options`,
+where `options` is `key=value` pairs separated by commas (`,`).  For example,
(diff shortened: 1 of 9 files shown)


openjdk/mmtkHeap.cpp
```diff
@@ -90,6 +90,10 @@ jint MMTkHeap::initialize() {
     bool set_options = process_bulk(strdup(ThirdPartyHeapOptions));
     guarantee(set_options, "Failed to set MMTk options. Please check if the options are valid: %s\n", ThirdPartyHeapOptions);
   }
+
+  // Read MMTk options from environment variables (such as `MMTK_THREADS`), overriding cmdline options.
+  mmtk_builder_read_env_var_settings();
```
qinsoon, line 95: The bug in #244 is that the env var value got overwritten by the command line value. But with the change, the command line value will get overwritten by the env var value. So the PR does not solve the issue, it just changes the priority of the option value from different sources.

We should not mix the use of env vars and command line options. Maybe it is time to stop using env vars. It may take some time for the scripts and running configs to keep up with the change. I guess we could make it a goal for the next release. We would have 6 weeks to fix broken stuff.

https://github.com/mmtk/mmtk-core/pull/955 is good. It allows us to disable setting options by env var easily. We could also accept https://github.com/mmtk/mmtk-core/pull/954. It is a workaround for the immediate issue #244. In the long run, it is also okay to allow users to check if an option is explicitly set or not.

qinsoon: In short, my suggestion is we review both mmtk-core PRs, and we move away from env var in the next release.

wenyuzhao: Actually I believe this solves the issue. For openjdk, there are three ways to set options: (1) `MMTK_` env vars (2) ThirdPartyHeapOptions (3) Copied from hotspot cmd args

The problem is that we're ignoring env var values and unconditionally using openjdk cmd arg for the `threads` option. This should be solved as long as we define a clear priority between these three sources, and make sure the one with higher priority can overwrite others, instead of ignoring it.

k-sareen: Tbh I'm not sure why you can't set `ParallelGCThreads` for OpenJDK instead of `MMTK_THREADS` in #242. Is there a particular reason we need to support that?

qinsoon: > Actually I believe this solves the issue. For openjdk, there are three ways to set options: (1) `MMTK_` env vars (2) ThirdPartyHeapOptions (3) Copied from hotspot cmd args
> 
> The problem is that we're ignoring env var values and unconditionally using openjdk cmd arg for the `threads` option. This should be solved as long as we define a clear priority between these three sources, and make sure the one with higher priority can overwrite others, instead of ignoring it.

OK. I guess the actual issue is that command line args always have a default value, so when we prioritise command line args, it always overwrites values from other sources. But if we prioritise env vars, it could be just empty if not set. If you think https://github.com/mmtk/mmtk-core/pull/954 and the OpenJDK PR is not needed, you can close them.

wks:
I figured out that we can use `FLAG_IS_DEFAULT(ParallelGCThreads)` to detect if the user explicitly set the `ParallelGCThreads` option.  I have adjusted this PR so that command line options (`-XX:ParallelGCThreads=xxxxx`) take priority over environment variables (`MMTK_THREADS=yyyyy`).

openjdk/mmtkHeap.cpp
wenyuzhao: > Tbh I'm not sure why you can't set ParallelGCThreads for OpenJDK instead of MMTK_THREADS in https://github.com/mmtk/mmtk-openjdk/issues/242. Is there a particular reason we need to support that?

`ParallelGCThreads` should work. The problem is we can only set `ParallelGCThreads` now. The binding will ignore the values in `MMTK_THREADS`

wks: > OK. I guess the actual issue is that command line args always have a default value, so when we prioritise command line args, it always overwrites values from other sources. But if we prioritise env vars, it could be just empty if not set.

Yes.  That's the cause.  But OpenJDK actually provides a way to tell whether a command line option is explicitly given or omitted as default (`FLAG_IS_DEFAULT(ParallelGCThreads)`).  I have adjusted the PR so that environment variables overrides the *default* of `ParallelGCThreads`, but if `-XX:ParallelGCThreads=xxxx` is given on the command line, it will override the environment variable instead.

> If you think [mmtk/mmtk-core#954](https://github.com/mmtk/mmtk-core/pull/954) and the OpenJDK PR is not needed, you can close them.

But after a second thought, https://github.com/mmtk/mmtk-core/pull/954 may be useful if we want to optionally pass non-default MMTk options to other places just like we optionally pass non-default OpenJDK options to MMTk.  We can keep both https://github.com/mmtk/mmtk-core/pull/954 and https://github.com/mmtk/mmtk-core/pull/955.

k-sareen: > > Tbh I'm not sure why you can't set ParallelGCThreads for OpenJDK instead of MMTK_THREADS in #242. Is there a particular reason we need to support that?
> 
> `ParallelGCThreads` should work. The problem is we can only set `ParallelGCThreads` now. The binding will ignore the values in `MMTK_THREADS`

That's what I'm saying, I'm not sure what the problem is with using `ParallelGCThreads`? Shouldn't we try to be as similar to the VM as possible? `MMTK_THREADS` and other environment variables are (still) there to quickly bootstrap new bindings/ports so that they don't have to bother fiddling with parsing commandline options.

openjdk/mmtkHeap.cpp
```diff
@@ -83,13 +83,25 @@ jint MMTkHeap::initialize() {
   const size_t max_heap_size = collector_policy()->max_heap_byte_size();
   //  printf("policy max heap size %zu, min heap size %zu\n", heap_size, collector_policy()->min_heap_byte_size());
 
-  // Set options
-  mmtk_builder_set_threads(ParallelGCThreads);
-  mmtk_builder_set_transparent_hugepages(UseTransparentHugePages);
+  // Read MMTk options from environment variables (such as `MMTK_THREADS`).
+  // i.e. Environment variable options (priority 3) override OpenJDK's default options (priority 2).
```
k-sareen, line 87: This priority should be documented somewhere. Either in the `README` or user guide.

wks: Done.  I documented the priorities of different option sources in openjdk/thirdPartyHeapArguments.cpp

openjdk/thirdPartyHeapArguments.cpp
```diff
@@ -50,7 +50,19 @@ void ThirdPartyHeapArguments::initialize() {
...
+  // 2. Default values defined in this function
+  // 3. Environment variables starting with `MMTK_`
+  // 4. Command line arguments
+  // We need to be careful about the order in which we set the options in the MMTKBuilder so that
+  // the values from the highest priority source will take effect.
+
+  // Set options in MMTKBuilder to OpenJDK's default options.
+  // i.e. OpenJDK's default options (priority 2) override MMTk's default options (priority 1).
+  mmtk_builder_set_threads(ParallelGCThreads);
+  mmtk_builder_set_transparent_hugepages(UseTransparentHugePages);
```
wenyuzhao, line 65: Probably move this to mmtkHeap.cpp before L89, so it is easier to infer the priority from ~10 lins of consecutive code.

wks: Done.  I moved them to mmtkHeap.cpp, and I further merged the code for Priority 2 and Priority 4 into one function, where a boolean argument controls whether to set default or overridden options.

wenyuzhao (approved):

README.md
```diff
@@ -255,4 +285,25 @@ Using scaled threading model. 24 processors detected, 24 threads used to drive t
...
+
+MMTk has many options defined in https://github.com/mmtk/mmtk-core/blob/master/src/util/options.rs
+
+You can use environment variables started with `MMTK_` to set those options.  For example,
+`export MMTK_THREADS=1` will set the number of GC worker threads to one.  Follow the link above for
+more details.
+
+You can also set those options via command line arguments.
+
+-   `-XX:ParallelGCThreads=n` (where `n` is a number) sets the number of GC worker threads.
```
caizixian, line 298: Could you clarify that this option actually changes both the number of STW/parallel GC threads and the concurrent GC threads for MMTk?

For OpenJDK GCs, `-XX:ParallelGCThreads=n` only changes the number of STW threads IIRC.

caizixian (changes requested):

README.md
k-sareen: Currently concurrent GC has not been merged into mmtk-core. So when Wenyu merges his branch, we can introduce the same interface (no of concurrent GC threads) to make it similar to the OpenJDK interface

wks: I added clarification about `-XX:ConcGCThreads`.  Because we haven't merged concurrent plans, yet, I just mentioned that `ConcGCThreads` is ignored when using MMTk.  When we merge concurrent plans, we can update this line of documentation, and maybe also adjust the MMTk options (such as introducing `Options::concurrent_threads`, etc.) that matches the OpenJDK options.

caizixian (approved):

PR #250, opened 2023-09-29
Bump version to v0.20

qinsoon:


caizixian (approved):
Same as mmtk/mmtk-jikesrvm. We should change the repo to only allow squashed merges.

qinsoon:
> Same as mmtk/mmtk-jikesrvm. We should change the repo to only allow squashed merges.

Yes. I have changed the settings, and only squash merge is allowed on our repos (mmtk-core, and mmtk-X).

PR #252, opened 2023-10-16
Update README about exploded build

qinsoon:
We should ask people to avoid using exploded image. This PR rephrases some related discussion about exploded image to warn people not to use exploded image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,12 @@ $ make CONF=linux-x86_64-normal-server-$DEBUG_LEVEL THIRD_PARTY_HEAP=$PWD/../mmt
 
 The output jdk is at `./build/linux-x86_64-normal-server-$DEBUG_LEVEL/jdk`.
 
-**Note:** The above `make` command will build what is known as the [`default` target or "exploded image"](https://github.com/openjdk/jdk11u/blob/master/doc/building.md#Running-make). This build is exclusively meant for developers who want quick and incremental builds to test changes. If you are planning on evaluating your build (be it performance, minimum heap, etc.), then it is *highly advised* to use the `images` target. The `default` target is the (roughly) minimal set of outputs required to run the built JDK and is not guaranteed to run all benchmarks. It may have bloated minimum heap values as well. The `images` target can be built like so:
+**Warning:** The above `make` command will build what is known as the [`default` target or "exploded image"](https://github.com/openjdk/jdk11u/blob/master/doc/building.md#Running-make).
+This build is **exclusively** meant for developers who want quick and incremental builds to test changes.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -118,7 +118,12 @@ $ make CONF=linux-x86_64-normal-server-$DEBUG_LEVEL THIRD_PARTY_HEAP=$PWD/../mmt
 
 The output jdk is at `./build/linux-x86_64-normal-server-$DEBUG_LEVEL/jdk`.
 
-**Note:** The above `make` command will build what is known as the [`default` target or "exploded image"](https://github.com/openjdk/jdk11u/blob/master/doc/building.md#Running-make). This build is exclusively meant for developers who want quick and incremental builds to test changes. If you are planning on evaluating your build (be it performance, minimum heap, etc.), then it is *highly advised* to use the `images` target. The `default` target is the (roughly) minimal set of outputs required to run the built JDK and is not guaranteed to run all benchmarks. It may have bloated minimum heap values as well. The `images` target can be built like so:
+**Warning:** The above `make` command will build what is known as the [`default` target or "exploded image"](https://github.com/openjdk/jdk11u/blob/master/doc/building.md#Running-make).
```
caizixian, line 121: I think the discussion from today's meeting is that we should change line 116 to use `images`, and then change 119 to `$CONF/images/jdk`. This makes it very clear that this is the default way of doing things.

And then down below, we add warning saying that if you omit `images`, it still compiles but it produces what's known as an exploded build, and comes with *bad consequences*.

cc @steveblackburn

caizixian (changes requested):

README.md
qinsoon: Angus has a PR for that change: https://github.com/mmtk/mmtk-openjdk/pull/253. Feel free to close this PR.

caizixian:
Closing in favor of #253

README.md
steveblackburn: @caizixian That's roughly what I was saying, yes.   My intention is that anyone reading the instructions will:
1) Be shown how to do so using `images`, and (implicitly) learn that this is "the way" to build.
2) Also discover (via a "footnote" of sorts), that the consequences of using the alternative (an exploded build) is dangerous, and comes with subtle *but consequential* side effects.

steveblackburn: My view was that 1. is essential, and that omitting 2. would be a mistake because some people might not realize what trouble they're buying into by trying the "logical alternative".

caizixian: Angus has an alternative PR that seems to match our discussion better https://github.com/mmtk/mmtk-openjdk/pull/253

PR #253, opened 2023-10-16
Change README to make images target the default build command

angussidney:
Also change the language to make it clearer that exploded images must not be used for performance builds.

Discussion question: is it worth making it clearer that PGO builds should be used for performance analysis?


caizixian:
>  is it worth making it clearer that PGO builds should be used for performance analysis?

I think so

k-sareen (approved):
LGTM. Thanks!

PR #255, opened 2023-10-18
Update to mmtk-core PR #988

qinsoon:
This PR updates mmtk-core to https://github.com/mmtk/mmtk-core/pull/988. Once https://github.com/mmtk/mmtk-core/pull/988 is merged, this PR should be updated to the new mmtk-core commit, and will be auto merged (if all the requirements for auto merging is met).


wks (approved):
LGTM.  Looks like this is only a test.

PR #257, opened 2023-11-02
Bump version to v0.21.0

wks:


qinsoon:
The binding should point to the mmtk-core PR. In that case, if there is any change in mmtk-core, this PR will be tested with the changes. There is no actual change in the mmtk-core PR so it doesn't matter this time.

qinsoon (approved):
LGTM

PR #258, opened 2023-11-18
Add Java-specific constants from MMTk

qinsoon:
Java-specific constants in MMTk are no longer public in https://github.com/mmtk/mmtk-core/pull/1026. This PR duplicates the constants that are used in the binding.


wks (approved):
LGTM

PR #263, opened 2024-01-08
Refactor CI test scripts

qinsoon:
This PR tries to consolidate the new dacapo chopin tests and the old dacapo 2006 tests to make them consistent. It also exposes tests to mmtk-core using Github workflows rather than shell scripts (WIP, https://github.com/mmtk/mmtk-core/issues/935). It fixes the heap sizes for 2006 tests (https://github.com/mmtk/mmtk-core/issues/424).

Changes in the PR:
* Separate building from testing when we run dacapo 2006 tests mmtk-openjdk CI.
* Using moderate heap sizes (proportional to the min heap size) to run dacapo 2006 tests.
* The old `ci-test.sh` still works (binding tests in the mmtk-core repo is still functioning).

WIP:
* `test-core-extended.yml` and `test-core-minimal.yml`  (https://github.com/mmtk/mmtk-core/issues/935) needs more work. However, I cannot test them at the moment. Workflow dispatch only works if the workflows exist in the default branch. I need to get them merged first, then start modifying and testing them.


.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+        fi
+    - name: Build MMTk OpenJDK ${{ inputs.debug-level }}
+      run: |
+        OPENJDK_PATH=$GITHUB_WORKSPACE/git/openjdk DEBUG_LEVEL=${{ inputs.debug-level }} OPENJDK_BUILD_TARGET=product-bundles ./.github/scripts/ci-build.sh
+      working-directory: ./git/mmtk-openjdk
+    - name: Upload bundles
+      uses: actions/upload-artifact@v3
+      with:
+        name: linux-x86_64-server-${{ inputs.debug-level }}-bundles-${{ env.BUILD_SUFFIX }}
+        path: ./git/openjdk/build/linux-x86_64-normal-server-${{ inputs.debug-level }}/bundles/*_bin.tar.gz
```
wks, line 70: This is the only place the `_bin.tar.gz` file (instead of `_bin-debug.tar.gz`) is used.  If we can use wildcard here, we won't need to change the file name.

.github/workflows/stylecheck.yml
```diff
@@ -0,0 +1,39 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  style-check:
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v2
```
wks, line 16: `checkout@v2` is too old.  We should bump it to `v4`.  Since this PR is doing refactoring, it is a good chance to do a global search/replacement for GitHub actions like this.

qinsoon: I changed all the `checkout` uses to `v4`.

.github/workflows/build.yml
qinsoon: I removed the renaming in `ci-build.sh`, and added two paths (`*_bin.tar.gz` and `*_bin-debug.tar.gz`) here for matching and uploading artifacts. I will see if it works.

wks:
The log shows that the CI successfully uploaded `jdk-11.0.19-internal+0_linux-x64_bin-debug.tar.gz` as an artifact.

But `pmd` failed with the `Immix` plan for both the `fastdebug` and the `release` builds.

And the log shows it is still using DaCapo Chopin RC3 instead of the released version.

qinsoon:
> But `pmd` failed with the `Immix` plan for both the `fastdebug` and the `release` builds.

We have random failures in the dacapo chopin tests.

> And the log shows it is still using DaCapo Chopin RC3 instead of the released version.

Yeah. I don't plan to move to the release version in this PR.

wks (approved):
LGTM

.github/scripts/ci-build.sh
```diff
@@ -2,12 +2,15 @@ set -xe
...
-unset MMTK_PLAN
+ensure_env OPENJDK_PATH
 
-# To OpenJDK folder
-cd $OPENJDK_PATH
+# Use fastdebug if DEBUG_LEVEL is unset
+DEBUG_LEVEL=${DEBUG_LEVEL:="fastdebug"}
+
+# Build target. Could be empty, or product-bundles.
+build_target=$1
```
k-sareen, line 11: This variable is unused

.github/scripts/ci-test-assertions.sh
```diff
@@ -4,57 +4,53 @@ set -xe
...
+run_subset() {
+    heap_multiplier=$1
 
-cd $OPENJDK_PATH
+    runbms_dacapo2006_with_heap_multiplier fop $heap_multiplier
+    runbms_dacapo2006_with_heap_multiplier luindex $heap_multiplier
+}
 
 # -- SemiSpace --
 export MMTK_PLAN=SemiSpace
```
k-sareen, line 15: I thought the idea was to move away from using environment variables since OpenJDK has support for passing MMTk options

PR #277, opened 2024-05-24
Add option to put forwarding bits on the side

wks:
Added a feature to force using on-the-side forwarding bits.  This is for testing mmtk-core.  The CI script for minimal test is also extended to run several plans involving CopySpace and ImmixSpace on OpenJDK with side forwarding bits.

Related PR: https://github.com/mmtk/mmtk-core/pull/1138 The feature introduced in this PR can be used to check the correctness of that PR.


qinsoon (approved):
LGTM

Issue #278, opened 2024-05-24
Deadlock in `resume_mutator`


wks:
This is reproducible when running `lusearch` in DaCapo 2006.  It seems it is more reproducible if we use `taskset` to limit the number of processors to use.  With the following command, it is almost 100% reproducible.

```
MMTK_PLAN=SemiSpace MMTK_THREADS=1 taskset -c 0-3 /path/to/openjdk/build/linux-x86_64-normal-server-fastdebug/jdk/bin/java -XX:+UseThirdPartyHeap -server -XX:MetaspaceSize=500M -Xm{s,x}32M -jar dacapo-2006-10-MR2.jar lusearch
```

- `MMTK_THREADS=1` is not important.  It can be reproduced with any number of GC threads.
- taskset -c 0-3` seems to influence the probability to reproduce.  The smaller the number of cores, the more likely it is to reproduce.

The stack trace shows that there is one mutator and one GC thread waiting.

The mutator thread is triggering a GC using `GCRequester`, which in turn is waiting for the mutex in the `WorkerMonitor`.

```
#0  syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
#1  0x00007ffff50c35e4 in std::sys::pal::unix::futex::futex_wait () at library/std/src/sys/pal/unix/futex.rs:62
#2  std::sys::pal::unix::locks::futex_mutex::Mutex::lock_contended () at library/std/src/sys/pal/unix/locks/futex_mutex.rs:56
#3  0x00007ffff540dd08 in std::sys::pal::unix::locks::futex_mutex::Mutex::lock (self=0x7ffff004aed0)
    at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/std/src/sys/pal/unix/locks/futex_mutex.rs:28
#4  0x00007ffff54228b5 in std::sync::mutex::Mutex<mmtk::scheduler::worker_monitor::WorkerMonitorSync>::lock<mmtk::scheduler::worker_monitor::WorkerMonitorSync> (self=0x7ffff004aed0)
    at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/std/src/sync/mutex.rs:273
#5  0x00007ffff53dce9b in mmtk::scheduler::worker_monitor::WorkerMonitor::make_request (self=0x7ffff004aed0, goal=mmtk::scheduler::worker_goals::WorkerGoal::Gc) at src/scheduler/worker_monitor.rs:101
[...]

PR #279, opened 2024-05-27
Fix deadlock related to safepoint sync.

wks:
`gc_lock` now never does safepoint check.  We always explicitly enter safepoint via `ThreadBlockInVM`.

The VM companion thread now lets `VM_MMTkSTWOperation` (executed by the VM thread) transition the state to `_threads_resumed` instead of doing it after `VMThread::execute` returns.  This works around a problem where `VMThread::execute` continues to block the companion thread when the VM thread starts executing other VM operations.

Some minor style changes:

We consistently use the constant `Mutex::_no_safepoint_check_flag` instead of `true` when acquiring or waiting for the `Monitor`.

Slightly updated comments and logs.  Logging statements no longer include thread IDs because thread IDs can be turned on by a command line option.

### DIFF
--- a/openjdk/mmtkUpcalls.cpp
+++ b/openjdk/mmtkUpcalls.cpp
@@ -74,19 +74,19 @@ static void mmtk_resume_mutators(void *tls) {
 #endif
 
   // Note: we don't have to hold gc_lock to increment the counter.
-  // The increment has to be done before mutators can be resumed
-  // otherwise, mutators might see a stale value
+  // The increment has to be done before mutators can be resumed (from `block_for_gc` or yieldpoints).
+  // Otherwise, mutators might see an outdated start-the-world count.
   Atomic::inc(&mmtk_start_the_world_count);
+  log_debug(gc)("Incremented start_the_world counter to %zu.", Atomic::load(&mmtk_start_the_world_count));
 
-  log_debug(gc)("Requesting the VM to resume all mutators...");
+  log_debug(gc)("Requesting the companion thread to resume all mutators blocking on yieldpoints...");
   MMTkHeap::heap()->companion_thread()->request(MMTkVMCompanionThread::_threads_resumed, true);
-  log_debug(gc)("Mutators resumed. Now notify any mutators waiting for GC to finish...");
 
+  log_debug(gc)("Notifying mutators blocking on the start-the-world counter...");
   {
-    MutexLockerEx locker(MMTkHeap::heap()->gc_lock(), true);
+    MutexLockerEx locker(MMTkHeap::heap()->gc_lock(), Mutex::_no_safepoint_check_flag);
     MMTkHeap::heap()->gc_lock()->notify_all();
   }
-  log_debug(gc)("Mutators notified.");
 }
 
 static const int GC_THREAD_KIND_WORKER = 1;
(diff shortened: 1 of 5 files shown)


k-sareen:
Why was this never a problem when we had the coordinator thread? Also it'd be good to get @tianleq to take a look at this PR since he's spent a lot of time understanding the internal safepoint thread mechanism in HotSpot.

wks:
> Why was this never a problem when we had the coordinator thread?

When we had the coordinator thread, `GCRequester` contained the variables that record whether GC is triggered, and they  were protected by their own mutex lock and condvar.  Mutators triggered GC by acquiring that mutex and signalling that condvar; the coordinator thread only acquired that mutex and waited for that condvar after one GC has finished and before the next GC starts.  Particularly, the coordinator does not hold any lock in EndOfGC.

After removing the coordinator thread, we removed the mutex and condvar in `GCRequester`, and there is effectly only one mutex that synchornizes both the mutators and the coordinator.  That is the mutex in `WorkerMonitor`.  `WorkerMonitor` used to be an unmamed pair of `(Mutex<()>, Condvar)` created by the scheduler, and that was the same mutex and condvar that protected work buckets.  As part of the new synchronization mechanism, the last parked GC worker behaves like the old coordinator and executes what used to be done in EndOfGC.  It does so while holding the mutex in `WorkerMonitor` so that other GC workers will not unpark.  Mutators also needs to hold that mutex to set "goals" (such as "please do a GC") and notify GC workers to respond to the newly added "goal".  Now, you see, mutators cannot trigger GC when the last parked worker is still executing EndOfGC (now named `GCWorkScheduler::on_gc_finished`).  That makes sense because it doesn't make sense to trigger GC before the last GC officially finished.  And even if a mutator runs so fast that it tried to trigger GC, the last parked GC worker will eventually finish and release the lock.  Actually, `resume_mutators` is the last thing `on_gc_finished` does, and after that the worker will release the mutex.

However, the synchronization problem with OpenJDK's VMThread blocked the "companion thread", which in turn blocked the worker executing `resume_mutators`.  That will block mutators that attempted to request GC, and that blocks the next VM operation that attempts to stop the world, and that's what was blocking the companion thread in the first place.  Deadlock.

So the reason why this problem only manifests after we removed the coordinator thread is that our synchronization mechanism in mmtk-core changed.

This problem is very likely to reproduce when executing `lusearch` in DaCapo 2006, but we never run that benchmark in CI.  

One reason why `lusearch` in DaCapo 2006 is likely to trigger the problem is that it executes an awful lot of `RevokeBias` VM operation (200+, evenly distributed during the benchmark, when given a 32M heap).  It doesn't execute any `RevokeBias` in `antlr`, only one in `pmd`, and only three in `luindex` (all from DaCapo 2006).  In DaCapo Bach, `lusearch executes 55 RevokeBias, The `luserach` in DaCapo Chopin is interesting.  It executes 200+ `RevokeBias` in the first 3 seconds when the benchmark is being prepared, and 5 `RevokeBias` at the end of the benchmark.  But during the 50 seconds of the main part of the benchmark, only one `RevokeBias` is executed.  I think the `lusearch` benchmark may have changed its behavior a lot during the course of 2006, Bach (2009) and Chopin (2023).  Since it doesn't appear often in the modern `lusearch`, we didn't notice that was a problem.

openjdk/mmtkVMCompanionThread.cpp
```diff
@@ -123,23 +107,70 @@ void MMTkVMCompanionThread::wait_for_reached(stw_state desired_state) {
...
+    // - The VM thread is blocked in `SafepointSynchronize::begin()`, waiting for all mutators to
+    //   reach safepoints.
+    // - One mutator is allocating too fast and triggers a GC, which requires the `WorkerMonitor`
+    //   lock in mmtk-core.
+    // - A GC worker is still executing `mmtk_resume_mutator`, holding the `WorkerMutator` (as the
+    //   last parked GC worker).  It is asking the companion thread to resume mutators, and is still
+    //   waiting for the companion thread to reach the `_thread_resumed` state.  As we see before,
+    //   the companion thread is waiting, too.
+    //
+    // By notifying now, we let the companion thread stop waiting, and therefore allowing the last
```
tianleq, line 155: I feel that this comment is confusing. If I understand correctly, the companion thread is not waiting on `_lock`, instead, it is waiting for the `VMThread` to notify it that the vm operation has finished. So the companion thread still needs to wait until the batch of vm operations to finish. However, because now gc threads no longer relies on the companion thread, they will be notified now deadlock no longer occurs.

 All other part looks good and makes perfect sense.

wks: You are right.  The comment is wrong.  The companion thread is not waiting on the `MMTkVMCompanionThread::_lock`, but on the `VMThread::VMOperationRequest_lock` while executing `VMThread::execute`.  If the VM thread notifies the `_lock` now, the companion thread will not stop waiting, but it will unblock the GC worker thread (running `resume_mutators`) which is waiting on `_lock` (in `MMTkVMCompanionThread::request`).

I'll rephrase the comment to "By notifying now, we unblock the GC worker waiting on `_lock` before the companion thread returns from `VMThread::execute`, and therefore allowing...".

tianleq (approved):
LGTM

openjdk/mmtkUpcalls.cpp
```diff
@@ -111,30 +111,37 @@ static void mmtk_spawn_gc_thread(void* tls, int kind, void* ctx) {
...
   size_t next_count = my_count + 1;
 
+  log_debug(gc)("Will block until the start_the_world counter reaches %zu.", next_count);
+
   {
-    // Once this thread acquires the lock, the VM will consider this thread to be "in safe point".
-    MutexLocker locker(MMTkHeap::heap()->gc_lock());
+    // Enter safepoint.
+    JavaThread* thread = JavaThread::current();
+    ThreadBlockInVM tbivm(thread);
```
wenyuzhao, line 133: I see `ThreadBlockInVM` makes the current thread stack walkable. But how can our previous implementation work without this guard?

wenyuzhao (approved):

openjdk/mmtkUpcalls.cpp
wks: Previously, `MutexLocker` does "safepoint check".  That is, if safepoint synchronization is in progress, it will yield when attempting to acquire the lock (i.e. there is a `ThhreadBlockInVM` hidden in `Monitor::lock` in the code path of yielding).  That worked, err... to a certain degree.

It looks like if the thread that tries to acquire the lock yields, other threads cannot acquire the same lock, either.  That prevented `release_mutator` from acquiring the `gc_lock`.  That was the cause of the first source of deadlock described [here](https://github.com/mmtk/mmtk-openjdk/issues/278#issuecomment-2129928182)  We didn't have this problem before we removed the coordinator thread.  I think that was because when a resumed mutator runs so fast that it triggers GC, it acquires a different lock, and that won't try to acquire the `WorkerMonitor` currently held by `release_mutator`.  When another VM operation, say RevokeBias, triggers another STW, the Java threads simply yield, and the VM threads do the VM operation and release the Java threads.  No deadlock.  The deadlock only happens if the mutator is blocked on the `WorkerMonitor` in `GCRequester::request` which is *not* executed in saferegion.

PR #282, opened 2024-07-08
Generational code cache roots

wks:
This PR ports the generational code cache roots handling code from the LXR branch. See https://github.com/wenyuzhao/mmtk-openjdk/blob/lxr/mmtk/src/gc_work.rs#L242-L262

We split code cache roots into two hash tables, one for nursery roots and one for mature roots.  Nursery roots are those added since the previous GC.  In a nursery GC, we only scan the nursery code cache roots, assuming mature code cache roots all point to mature objects.  In a full-heap GC, we scan both nursery and mature code cache roots.

Because code cache roots are added in bulk in the beginning of execution, and are seldom added later, we expect the number of code cache roots to reduce to zero for most nursery GCs, and greatly reduce the root-scanning time.

This PR also adds USDT tracepoints for code cache roots, and make use of the extensible eBPF timeline tools in mmtk-core introduced in https://github.com/mmtk/mmtk-core/pull/1162.

### DIFF
--- a/mmtk/src/api.rs
+++ b/mmtk/src/api.rs
@@ -497,7 +497,7 @@ thread_local! {
 /// Report a list of pointers in nmethod to mmtk.
 #[no_mangle]
 pub extern "C" fn mmtk_add_nmethod_oop(addr: Address) {
-    NMETHOD_SLOTS.with(|x| x.borrow_mut().push(addr))
+    NMETHOD_SLOTS.with_borrow_mut(|x| x.push(addr))
 }
 
 /// Register a nmethod.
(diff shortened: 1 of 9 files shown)


mmtk/src/api.rs
```diff
@@ -490,49 +490,41 @@ pub extern "C" fn get_finalized_object() -> NullableObjectReference {
...
-    let mut roots = crate::CODE_CACHE_ROOTS.lock().unwrap();
-    if let Some(slots) = roots.remove(&nm) {
-        // Relaxed sub instead of `fetch_sub`, since we've already acquired the lock.
-        crate::CODE_CACHE_ROOTS_SIZE.store(
-            crate::CODE_CACHE_ROOTS_SIZE.load(Ordering::Relaxed) - slots.len(),
-            Ordering::Relaxed,
-        );
+    {
+        let mut roots = crate::NURSERY_CODE_CACHE_ROOTS.lock().unwrap();
+        roots.remove(&nm);
```
qinsoon, line 524: Is it possible that `nm` can be found in both the nursery and the mature code caches? It is alright if you just want to make sure that it is removed from either the mature or the nursery cache.

qinsoon (approved):
LGTM. Probably merge after https://github.com/mmtk/mmtk-openjdk/pull/288, so we can make sure the CI tests pass for this PR.

mmtk/src/api.rs
wks: If the C++ side never registers the same nmethod twice before it is unregistered, it will be impossible to find the same nmethod in both nursery and mature caches.  The code only ensures we remove it from both caches because depending on whether a GC happens between registering and unregistering the nmethod, it may be found in either of them.

PR #288, opened 2024-09-10
Bump version of action/upload-artifact

qinsoon:
https://github.com/mmtk/mmtk-openjdk/pull/287 bumps the version of download-artifact. We need to use a compatible version for upload-artifact.


wks (approved):
LGTM

PR #292, opened 2024-11-27
Add tests with large heap size

qinsoon:


wks (approved):
LGTM, but we should merge this after https://github.com/mmtk/mmtk-openjdk/pull/293 because these tests will fail without that fix.

Issue #291, opened 2024-11-25
ObjectReference is required to be word aligned.  addr: 0x8000026


qinsoon:
Reproduciable with v0.29.0. It is a bit strange that I didn't see the issue before. It would be good if anyone can try reproduce it and see if the issue actually exists.

MMTk OpenJDK: 1fcc9f10ae6a46b0cc0c78d7f11f19ca6d28e566 (v0.29.0)
OpenJDK: 28e56ee32525c32c5a88391d0b01f24e5cd16c0f (as recorded in `Cargo.toml` with the biding commit)

Build:
```
export DEBUG_LEVEL=slowdebug && sh configure --disable-warnings-as-errors --with-debug-level=$DEBUG_LEVEL && make CONF=linux-x86_64-normal-server-$DEBUG_LEVEL THIRD_PARTY_HEAP=/home/yilin/Code/openjdk_workspace/mmtk-openjdk/openjdk
```

Run:
```
yilin@rat:~/Code/openjdk_workspace/openjdk$ RUST_BACKTRACE=1 build/linux-x86_64-normal-server-slowdebug/jdk/bin/java -XX:+UseThirdPartyHeap -jar /usr/share/benchmarks/dacapo/dacapo-23.11-chopin.jar fop
[2024-11-25T06:20:07Z INFO  mmtk::memory_manager] Initialized MMTk with GenImmix (DynamicHeapSize(6815744, 8391753728))
[2024-11-25T06:20:07Z INFO  mmtk::util::heap::gc_trigger] [POLL] nursery: Triggering collection (1670/1664 pages)
[2024-11-25T06:20:07Z INFO  mmtk::plan::generational::global] Nursery GC
[2024-11-25T06:20:07Z INFO  mmtk::plan::generational::immix::global] Nursery GC
[2024-11-25T06:20:07Z WARN  mmtk::util::heap::gc_trigger] Proportional nursery with min size 0.25 (1703936) is smaller than DEFAULT_MIN_NURSERY (2097152). Use DEFAULT_MIN_NURSERY instead.
[2024-11-25T06:20:07Z INFO  mmtk::scheduler::scheduler] End of GC (276/1664 pages, took 9 ms)
thread '<unnamed>' panicked at /home/yilin/.cargo/git/checkouts/mmtk-core-3306bdeb8eb4322b/8640ab8/src/util/address.rs:589:9:
ObjectReference is required to be word aligned.  addr: 0x817606c
stack backtrace:
[...]

PR #294, opened 2024-12-04
Skip val register in ObjectBarrier impl

wks:
`MMTkBarrierSetAssembler::store_at` calls `object_reference_write_post` after calling `BarrierSetAssembler::store_at`.  However, `BarrierSetAssembler::store_at` modifies the `val` register to compress its value in place.  Consequently, `object_reference_write_post` will see a compressed pointer in the `val` register.

This PR makes two changes.

Firstly, in `MMTkObjectBarrierSetAssembler::object_reference_write_post`, we simply set both `c_rarg1` and `c_rarg2` to 0 before calling the write barrier slow path.  We exploit the semantics of the `ObjectBarrier` that it simply logs the object without looking at the slot or the target.  This will fix the [assertion error](https://github.com/mmtk/mmtk-openjdk/issues/291) because 0 will be interpreted as a `None` of type `Option<ObjectReference>`.

Secondly, we add a `bool compensate_val_reg` parameter to  `MMTkBarrierSetAssembler::object_reference_write_post` so that if we call it after `BarrierSetAssembler::store_at`, we can give `object_reference_write_post` a chance to decompress the compressed oop in the `val`.  This is intended for implementing *other barriers introduced in the future* that may use the `val` register, and keep the developers informed that the `val` register is mutated in `BarrierSetAssembler::store_at`.

Fixes: https://github.com/mmtk/mmtk-openjdk/issues/291
Related PR: https://github.com/mmtk/mmtk-openjdk/pull/293

### DIFF
--- a/openjdk/barriers/mmtkObjectBarrier.cpp
+++ b/openjdk/barriers/mmtkObjectBarrier.cpp
@@ -38,8 +38,11 @@ void MMTkObjectBarrierSetRuntime::object_reference_write_post(oop src, oop* slot
 
 #define __ masm->
 
-void MMTkObjectBarrierSetAssembler::object_reference_write_post(MacroAssembler* masm, DecoratorSet decorators, Address dst, Register val, Register tmp1, Register tmp2) const {
+void MMTkObjectBarrierSetAssembler::object_reference_write_post(MacroAssembler* masm, DecoratorSet decorators, Address dst, Register val, Register tmp1, Register tmp2, bool compensate_val_reg) const {
   if (can_remove_barrier(decorators, val, /* skip_const_null */ true)) return;
+
+  bool is_not_null = (decorators & IS_NOT_NULL) != 0;
+
   Register obj = dst.base();
 #if MMTK_ENABLE_BARRIER_FASTPATH
   Label done;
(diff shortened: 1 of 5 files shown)


openjdk/barriers/mmtkObjectBarrier.cpp
```diff
@@ -67,17 +70,21 @@ void MMTkObjectBarrierSetAssembler::object_reference_write_post(MacroAssembler*
...
 
   __ movptr(c_rarg0, obj);
-  __ lea(c_rarg1, dst);
-  __ movptr(c_rarg2, val == noreg ?  (int32_t) NULL_WORD : val);
-  __ call_VM_leaf_base(FN_ADDR(MMTkBarrierSetRuntime::object_reference_write_slow_call), 3);
+  __ xorptr(c_rarg1, c_rarg1);
+  // Note: If `compensate_val_reg == true && UseCompressedOops === true`, the `val` register will be
+  // holding a compressed pointer to the target object. If the write barrier needs to know the
+  // target, we will need to decompress it before passing it to the barrier slow path. However,
+  // since we know the semantics of `ObjectReference`, i.e. it logs the object without looking at
```
qinsoon, line 80: Did you mean "the semantics of object remembering barrier"?

We need to a check to know this is object remembering barrier somewhere in the code. We can get it from https://github.com/mmtk/mmtk-core/blob/a75309373cd301acad473bbb8ea04670facd6e2a/src/plan/plan_constraints.rs#L35, and store the value somewhere for the C++ code to access.

wks: > Did you mean "the semantics of object remembering barrier"?

Oops. That's a typo.  I did mean `ObjectBarrier`.

> We need to a check to know this is object remembering barrier somewhere in the code.

The changed code is in `mmtkObjectBarrier.cpp` which is dedicated to the object remembering barrier.  The class `MMTkObjectBarrierSetAssembler` is only used if `get_selected_barrier()` in `mmtkBarrierSet.cpp` returns `new MMTkObjectBarrier();`  Therefore we don't need to check here, as long as the `get_selected_barrier()` chooses the right barrier.

qinsoon (approved):

Issue #262, opened 2023-12-21
Validation FAILED for pmd default (GenImmix/StickyImmix)


qinsoon:
In the run: https://github.com/mmtk/mmtk-openjdk/actions/runs/7282018250/job/19843937295?pr=261

```console
pmd 2500 728 0abc..fg
Defaulting to user installation because normal site-packages is not writeable
Requirement already satisfied: pyyaml in /home/runner/.local/lib/python3.10/site-packages (6.0.1)
0abc..fg
Expected:
{'SemiSpace': 'pass', 'GenCopy': 'pass', 'Immix': 'pass', 'GenImmix': 'pass', 'StickyImmix': 'pass', 'MarkSweep': 'pass', 'MarkCompact': 'pass'}
Actual:
{'SemiSpace': 'pass', 'GenCopy': 'pass', 'Immix': 'pass', 'GenImmix': 'fail', 'StickyImmix': 'fail', 'MarkSweep': 'pass', 'MarkCompact': 'pass'}
=====
Expect GenImmix to pass, but it failed.
Expect StickyImmix to pass, but it failed.

Print logs for all failed runs: ['GenImmix', 'StickyImmix']

----------------------------------------------
START: /tmp/linux-x86_64-pmd-fastdebug-fv-az883-399-2023-[12](https://github.com/mmtk/mmtk-openjdk/actions/runs/7282018250/job/19843937295?pr=261#step:18:13)-21-Thu-000507/pmd.2500.728.jdk11-master.ms.s.fail_on_oom.tph.preserve.mmtk_gc-GenImmix.dacapo-23.9-RC3-chopin-ci.log.gz

-----
mkdir -p PLOTTY_WORKAROUND; timedrun; MMTK_PLAN=GenImmix /home/runner/work/mmtk-openjdk/mmtk-openjdk/bundles/jdk/bin/java -XX:MetaspaceSize=500M -XX:+DisableExplicitGC -server -XX:+CrashOnOutOfMemoryError -XX:+UseThirdPartyHeap -Xms728M -Xmx728M -jar /home/runner/work/mmtk-openjdk/mmtk-openjdk/dacapo/dacapo-23.9-RC3-chopin.jar -n 1 pmd -preserve
running-ng v0.4.4
Thu Dec 21 00:07:11 UTC 2023

 00:07:11 up 7 min,  0 users,  load average: 4.56, 2.80, 1.26
[...]

PR #298, opened 2025-03-18
Use buffer size exported from MMTk core

k-sareen:


caizixian (approved):

caizixian (dismissed):
Need to bump core version

k-sareen:
Ugh. I'm stupid. Put the constant in the wrong place.

k-sareen:
@qinsoon What is the protocol here? Does the MSRV for MMTk OpenJDK need to be bumped?

qinsoon:
> @qinsoon What is the protocol here? Does the MSRV for MMTk OpenJDK need to be bumped?

Yeah. The MSRV of the binding has to be the same or newer than that of mmtk-core.

k-sareen:
Is that a separate PR or the same PR?

qinsoon:
> Is that a separate PR or the same PR?

You can just do it in this PR.

k-sareen:
@qinsoon Can you approve this PR?

qinsoon (approved):

k-sareen:
This is really odd... It shouldn't fail since the PR doesn't change the semantics. The only difference is we've upgraded the MMTk core version to be more recent. @qinsoon what are your thoughts on this?

qinsoon:
> This is really odd... It shouldn't fail since the PR doesn't change the semantics. The only difference is we've upgraded the MMTk core version to be more recent. @qinsoon what are your thoughts on this?

No idea. Probably just rerun it first. We saw the same random fails on `pmd` for GenImmix and StickyImmix: https://github.com/mmtk/mmtk-openjdk/issues/262

k-sareen:
I've already re-ran it once

wks:
The pmd benchmark has been failing for a long time. https://github.com/dacapobench/dacapobench/issues/310

BTW, currently the thing that actually blocks this PR from being merged is @caizixian's request for change, not the PMD benchmark.

PR #300, opened 2025-04-17
Don't let VMThread acquire Heap_lock

wks:
The Heap_lock performs safepoint check.  Mutators may hold such locks and enter safepoint.  If the VMThread attempts to acquire this lock in a VMOperation while some mutators are still holding this lock, the VMThread will have to wait forever because mutators will not resume until the VMOperation ends.  Later versions of OpenJDK forbid the VMThread from acquiring any lock that performs safepoint check.

We notify the ReferenceHandler thread when resuming mutators.  But the current code does this notification in a VMOperation, and will cause the deadlock.  When this happens, the process will freeze when a GC is triggered.  This is likely to happen when the heap size is small and GC is triggered often.

We move the notification code to `mmtk_resume_mutators` which is executed on a GC worker thread *after* mutators have been resumed.  This solves the deadlock issue.

### DIFF
--- a/openjdk/mmtkUpcalls.cpp
+++ b/openjdk/mmtkUpcalls.cpp
@@ -87,6 +87,15 @@ static void mmtk_resume_mutators(void *tls) {
     MutexLockerEx locker(MMTkHeap::heap()->gc_lock(), Mutex::_no_safepoint_check_flag);
     MMTkHeap::heap()->gc_lock()->notify_all();
   }
+
+  log_debug(gc)("Notifying mutators blocking on Heap_lock for reference pending list...");
+  // Note: That's the ReferenceHandler thread.
+  {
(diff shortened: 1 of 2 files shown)


wks:
This change is required for porting to OpenJDK 21 anyway.  I made the same change to the `jdk-21+35-mmtk` branch because otherwise there will be an assertion error.  And I accidentally discovered this deadlock problem in the current `master` branch, and it is quite likely to trigger when running lusearch in DaCapo Chopin with a relatively small heap size.  (40M with StickyImmix is enough to trigger it most of the time).  And I wonder why it has been working well for so long.  Maybe because we usually test with -Xcomp, but the deadlock is the most likely to occur when the C1 JIT compiler holds this lock and attempt to execute a VMOperation.

qinsoon (approved):
LGTM

PR #305, opened 2025-04-23
(don't merge now) Fix CI tests for OpenJDK 21

wks:
This PR is for testing the CI scripts for OpenJDK 21.  Don't merge it now.


wks:
The good thing is that most test cases passed.  There is a Clippy warning about `enum` variant name in `KlassID`.  We should either rename `KlassID` to `KlassKind` because it was renamed upstream, or make it an alias of i32 because it is not used in the Rust part anyway.  And I think we should double check those mirrored data structures because other types may be renamed upstream, too.

We should also check if the CI scripts in mmtk-core, especially minimal test, work.

.github/configs/base.yml
```diff
@@ -83,19 +83,19 @@ plugins:
...
-  - "jdk11-master|ms|s|fail_on_oom|tph|preserve|mmtk_gc-StickyImmix"
-  - "jdk11-master|ms|s|fail_on_oom|tph|preserve|mmtk_gc-MarkSweep"
-  - "jdk11-master|ms|s|fail_on_oom|tph|preserve|mmtk_gc-MarkCompact"
+  - "jdk21-master|dacapochopin_jdk21|ms|s|fail_on_oom|tph|preserve|mmtk_gc-SemiSpace"
+  - "jdk21-master|dacapochopin_jdk21|ms|s|fail_on_oom|tph|preserve|mmtk_gc-GenCopy"
+  - "jdk21-master|dacapochopin_jdk21|ms|s|fail_on_oom|tph|preserve|mmtk_gc-Immix"
+  - "jdk21-master|dacapochopin_jdk21|ms|s|fail_on_oom|tph|preserve|mmtk_gc-GenImmix"
+  - "jdk21-master|dacapochopin_jdk21|ms|s|fail_on_oom|tph|preserve|mmtk_gc-StickyImmix"
+  - "jdk21-master|dacapochopin_jdk21|ms|s|fail_on_oom|tph|preserve|mmtk_gc-MarkSweep"
+  - "jdk21-master|dacapochopin_jdk21|ms|s|fail_on_oom|tph|preserve|mmtk_gc-MarkCompact"
```
wks, line 98: I think those modifiers expect the benchmark suite name to be exactly `dacapochopin`, but ours is `dacapo-23.11-MR2-chopin-ci`.  There should be a way to override it, or we can just replicate what's in `running-ng/src/running/config/base/dacapo.yml` to our own modifier here.

wks: I just renamed the benchmark suite to `dacapochopin` and it works.

wks:
Closed in favor for https://github.com/mmtk/mmtk-openjdk/pull/307

PR #306, opened 2025-04-25
Fix a typo in run_all_no_compressed_oop

wks:
This function should take one argument, but it used $2 instead. In previous runs, invocations of run_all_no_compressed_oop have failed silently without making the GitHub-level CI test case `ci-test-only-normal-no-compressed-oops` fail.

This PR fixes the typo.

### DIFF
--- a/.github/scripts/ci-test-only-normal-no-compressed-oops.sh
+++ b/.github/scripts/ci-test-only-normal-no-compressed-oops.sh
@@ -5,7 +5,7 @@ set -xe
 unset JAVA_TOOL_OPTIONS
 
 run_all_no_compressed_oop() {
-    heap_multiplier=$2
+    heap_multiplier=$1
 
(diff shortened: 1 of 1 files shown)


wks:
This is a strange bug.  Previously the test case about compressed oops have been "successful" so far, which should not happen.

Take https://github.com/mmtk/mmtk-openjdk/actions/runs/14508960198/job/40704006542#step:7:37 as an example.  The error `./.github/scripts/common.sh: line 31: -XX:-UseCompressedOops: syntax error in expression (error token is ":-UseCompressedOops")` appeared 7 times in that GitHub job, but the whole job was successful.

On the Bash level, if a syntax error happens in `$(( blah * blah))`, the enclosing bash function will return 1.  But even with `set -xe`, failure in one command won't automatically terminate the whole script.

This PR should fix this particular bug, but we should think about how to make our test scripts more robust against such failures.

qinsoon (approved):

PR #307, opened 2025-04-29
Merge the jdk-21+35-mmtk branch into master

wks:
We merge the branch for OpenJDK 21 into `master`.  From now on, the `master` branch of the `mmtk-openjdk` repository will directly target OpenJDK 21.  The `jdk-21+35-mmtk` branch will be decommissioned, and will remain its current state (i.e. the state just before merging into `master`), and will only be looked up when any developers are interested in the history of the development of the OpenJDK binding.

The following things work:

-   We can run OpenJDK 21 with MMTk on x86_64, and it supports compressed oops.
-   CI tests for pull requests of the `mmtk-openjdk` binding work.
-   CI binding tests (including both minimal tests and extended tests) for pull requests of the `mmtk-core` repo work.

One important thing that doesn't work immediately is the **performance regression CI**.  It requires a change in the `ci-perf-kit` repo and also needs related changes in the CI scripts in `mmtk-core`.  This PR is for OpenJDK, and will not trigger the performance regression CI.  This leaves us a small window to update `mmtk-core` and `ci-perf-kit` after this PR is merged.

**This PR should not be merged with "squash merge"**.  Instead we should make a proper merge commit so that the `jdk-21+35-mmtk` branch will be considered as an ancestor of the `master` branch from Git's point of view.  GitHub will also consider the branch `jdk-21+35-mmtk` as "merged".

# Major changes compared to the OpenJDK 11 binding

The OpenJDK repository is now https://github.com/mmtk/jdk.  We made this change to align with the naming convention of the repositories of the upstream.   The old repository https://github.com/mmtk/openjdk will be decommissioned.

We also made the following changes in the mmtk-openjdk repo:

-   Changes are made to reflect upstream changes in OpenJDK, including initialization, mutex API, roots, JIT compilers, etc.
-   We now support multiple architectures, including x86_64, ARM64 and RISC-V 64.  We changed the directory structure so that architecture-specific parts are held in separate files.
-   We updated the CI scripts to reflect related changes, such as checking out OpenJDK in the `jdk` directory instead of `openjdk`, and remove the `-normal` part from the OpenJDK CONF string.  We also slightly refactored the CI scripts to reduce the repetition of the path "repos/jdk" which used to be "repos/openjdk".


wks:
Closed in favor for https://github.com/mmtk/mmtk-openjdk/pull/335.  Branch jdk-21+35-mmtk lacks important fixes to be compiled with modern GCC and GLibC versions.

PR #312, opened 2025-08-04
Enable reference processing by default in CI

k-sareen:
Related to: https://github.com/mmtk/mmtk-core/pull/1360


k-sareen:
It's only failing currently because it's pointing to the default MMTk core which doesn't have the fix for reference processing in the Compressor yet.

qinsoon:
> It's only failing currently because it's pointing to the default MMTk core which doesn't have the fix for reference processing in the Compressor yet.

You can just modify `Cargo.toml` and point to the mmtk-core commit. Otherwise this PR is not testing anything new.

PR #325, opened 2025-08-20
Warn instead of panic when `object_iterate` is called.

wks:
The` MMTkHeap` instance will receive calls to `safe_object_iterate` when connected from VisualVM.  If we use `guarantee(...)` to indicate the function is not implemented, the process will abort.  We now give a warning instead.  This does not properly implement the iteration, but will allow the VM to be profiled with VisualVM.

### DIFF
--- a/openjdk/mmtkHeap.cpp
+++ b/openjdk/mmtkHeap.cpp
@@ -315,13 +315,13 @@ GrowableArray<MemoryPool*> MMTkHeap::memory_pools() {//may cause error
 
 // Iterate over all objects, calling "cl.do_object" on each.
 void MMTkHeap::object_iterate(ObjectClosure* cl) { //No need to implement.Traced whole path.Only other heaps call it.
-  guarantee(false, "object iterate not supported");
+  // Not supported.
 }
 
 // Similar to object_iterate() except iterates only
 // over live objects.
 void MMTkHeap::safe_object_iterate(ObjectClosure* cl) { //not sure..many dependencies from vm
-  guarantee(false, "safe object iterate not supported");
+  // Not supported.
 }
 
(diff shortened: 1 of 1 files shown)


wks:
One CI test failed with SIGSEGV, but I don't think it is related to this PR.

https://github.com/mmtk/mmtk-openjdk/actions/runs/17094762315/job/48477770783?pr=325

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f2c3b8136c8, pid=2200, tid=2208
#
# JRE version: OpenJDK Runtime Environment (11.0.19) (fastdebug build 11.0.19-internal+0-adhoc.runner.openjdk)
# Java VM: OpenJDK 64-Bit Server VM (fastdebug 11.0.19-internal+0-adhoc.runner.openjdk, mixed mode, tiered, compressed oops, third-party gc, linux-amd64)
# Problematic frame:
# C  [libmmtk_openjdk.so+0x4136c8]  mmtk_openjdk::abi::OopDesc::size::hf0ca8c7f40be73ed+0x18
#
# Core dump will be written. Default location: Core dumps may be processed with "/lib/systemd/systemd-coredump %P %u %g %s %t 9223372036854775808 %h %d" (or dumping to /home/runner/work/mmtk-openjdk/mmtk-openjdk/core.2200)
#
# An error report file with more information is saved as:
# /home/runner/work/mmtk-openjdk/mmtk-openjdk/hs_err_pid2200.log
#
# If you would like to submit a bug report, please visit:
#   https://bugreport.java.com/bugreport/crash.jsp
#
```

qinsoon (approved):

PR #327, opened 2025-08-29
Enable compressed pointers for Compressor

k-sareen:
MMTk core PR: https://github.com/mmtk/mmtk-core/pull/1372


qinsoon:
I am not sure why this PR did not get updated and auto merged with https://github.com/mmtk/mmtk-core/pull/1372.

PR #330, opened 2025-09-19
Reduce the test heap size for concurrent immix

qinsoon:
The heap size for concurrent immix was set to 7x, as initially it was non-moving. Later we allowed moving for the full pause in concurrent Immix. This PR reduces the heap size for our concurrent immix tests.

### DIFF
--- a/.github/configs/large-heap.yml
+++ b/.github/configs/large-heap.yml
@@ -1,9 +1,0 @@
-includes:
-  - "./base.yml"
-
-configs:
-  - "jdk11-master|ms|s|fail_on_oom|tph|preserve|mmtk_gc-ConcurrentImmix"
-
-# This will be expanded in CI when we run with the config. Keep a new line at the end.
-benchmarks:
-  dacapo-23.9-RC3-chopin-ci:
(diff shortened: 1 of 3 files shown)


wks:
> Wouldn't this just make it into a STW GC? Does it do any concurrent collections for 2.5x min heap?

This is a valid concern.  If the heap is too small (I mean having the same heap size as the STW Immix plan), it may end up triggering only STW GCs.  For example, when I run lusearch locally, if I set the heap size to 40MB or 80MB, ConcurrentImmix will trigger a few concurrent GC when the test harness is starting.  But when the formal benchmark starts, it only triggers full GCs.  But if I set the heap size to 120MB, it can run to the end, only triggering concurrent GCs.

So we may want to test Concurrent GC with two heap sizes, one small and one large.  Make sure both concurrent and full-heap GCs are triggered.  If that's too time-consuming, we can run all benchmarks with one heap size, and run lusearch (or another selected benchmark) with a different heap size.

qinsoon:
I checked two benchmarks: fop and lusearch. For fop, there are two concurrent GCs at the beginning, then it always does full GC for the rest. For lusearch, almost all the GCs are concurrent.

I think running with small heap sizes won't 'make it into a STW GC' -- more full GCs would be triggered if there is memory pressure, but this depends on the workload and the minimal heap size for the work load.

wks (approved):
LGTM

Issue #331, opened 2025-09-21
Duplicate edges for SemiSpace/luindex


qinsoon:
First seen in [run-dacapo-2006 / Test feature build (ci-test-assertions)](https://github.com/mmtk/mmtk-openjdk/actions/runs/17847477168/job/50750454830#logs)

```console
/home/runner/work/mmtk-openjdk/mmtk-openjdk/bundles/jdk/bin/java -XX:+UseThirdPartyHeap -server -XX:MetaspaceSize=100M -Xms24M -Xmx24M -jar /home/runner/work/mmtk-openjdk/mmtk-openjdk/dacapo/dacapo-2006-10-MR2.jar luindex
[2025-09-19T03:37:26Z INFO  mmtk::memory_manager] Initialized MMTk with SemiSpace (FixedHeapSize(25165824))
[2025-09-19T03:37:26Z WARN  mmtk::memory_manager] The feature 'extreme_assertions' is enabled. MMTk will run expensive run-time checks. Slow performance should be expected.
===== DaCapo luindex starting =====
...
adding luindex/william/README
adding luindex/william/comedies/allswellthatendswell
[2025-09-19T03:37:26Z INFO  mmtk::util::heap::gc_trigger] [POLL] copyspace0: Triggering collection (6149/6144 pages)
[2025-09-19T03:37:27Z INFO  mmtk::scheduler::scheduler] End of GC (1257/6144 pages, took 213 ms)
adding luindex/william/comedies/asyoulikeit
adding luindex/william/comedies/comedyoferrors
adding luindex/william/comedies/cymbeline
adding luindex/william/comedies/loveslabourslost
[2025-09-19T03:37:27Z INFO  mmtk::util::heap::gc_trigger] [POLL] copyspace1: Triggering collection (6157/6144 pages)
[2025-09-19T03:37:27Z INFO  mmtk::scheduler::scheduler] End of GC (1384/6144 pages, took 154 ms)
adding luindex/william/comedies/measureforemeasure
adding luindex/william/comedies/merchantofvenice
adding luindex/william/comedies/merrywivesofwindsor
adding luindex/william/comedies/midsummersnightsdream
[2025-09-19T03:37:28Z INFO  mmtk::util::heap::gc_trigger] [POLL] copyspace0: Triggering collection (6149/6144 pages)
[2025-09-19T03:37:28Z INFO  mmtk::scheduler::scheduler] End of GC (1480/6144 pages, took 192 ms)
[...]

PR #332, opened 2025-09-22
Refactor barrier implementations

wks:
We recently added another write barrier, the `SATBBarrier`.  Being another barrier based on the per-object unlog bit, it shares much code with the `ObjectBarrier`.  We refactor the write barriers to improve code sharing and clean up unused code.

We introduce abstract base classes `MMTkUnlogBitBarrierSetRuntime`, `MMTkUnlogBitBarrierSetAssembler`, `MMTkUnlogBitBarrierSetC1` and `MMTkUnlogBitBarrierSetC2`, which both `ObjectBarrier` and `SATBBarrier` derive from.  Those base classes include common parts related to unlog bit handling, including the fast paths for checking if the unlog bit is set, and also the generic implementation for both the pre and the post barriers if applicable.

Runtime:

- Extracted a method for testing if the unlog bit is set for a given object.

Assembler:

- Extracted a method for emitting the unlog bit checking fast path.
- Extracted a method for generating both the pre and the post write barriers.
- Simplified the fast path to use fewer instructions and fewer scratch registers.
- Worked around an issue where the `tmp1` and `tmp2` temporary registers passed in from `BarrierSetAssembler::store_at` may overlap with `dst.base()` and `dst_index()`, respectively.
  - See https://bugs.openjdk.org/browse/JDK-8301371

C1:

- Changed "runtime stubs" to be one per runtime function, not one per pre/post/refload barrier.  Those "runtime stubs" are now assembled by `MMTkBarrierSetAssembler` instead of by `MMTkObjectBarrierSetAssembler` or `MMTkSATBBarrierSetAssembler`.  The intention is that all barriers can call those functions.
- Extracted a method for emitting the unlog bit checking fast path.
- Extracted a method for emitting both the pre and the post write barriers and their slow-path code stub.
- We no longer pass the `slot` and `new_val` to the runtime function in the slow path.  Object-logging barriers only needs the `obj` argument for scanning its fields.
  - For this reason, we moved the C1 CodeStub for the pre/post barrier slow paths down from `mmtkBarrierSetC1.hpp` to `mmtkUnlogBitBarrier.hpp` (and unified the pre/post stub implementation) because it no longer needs the `slot` or `new_val` args, and is now specific to unlog bit barriers.  If we introduce field-logging barriers, we will implement a different stub.
- The C1 SATBBarrier no longer needs code patching.  The code patching was intended for barriers that load from the field in the fast path using LIR instructions, such as G1's SATB barrier.  Our object-logging barrier scans the fields in runtime functions (`mmtk_object_reference_write_pre`), not using LIR instructions.

C2:

- Extracted a method for emitting the unlog bit checking fast path.
- Extracted a method for emitting both the pre and the post write barriers.
- We no longer pass the `slot` and `new_val` to the runtime function in the slow path.

Misc:

- Removed the `#define private public` hack as we no longer need it.
- Removed unused macros, dead code and irrelevant comments.
- Changed the macro `SOFT_REFERENCE_LOAD_BARRIER` to a run-time flag `mmtk_enable_reference_load_barrier` settable by environment variable.


k-sareen:
I think we should specify in the type that this is an object remembering barrier. LXR's field remembering barrier would also be an "unlog-bit" barrier but it's a different implementation.

wks:
> I think we should specify in the type that this is an object remembering barrier. LXR's field remembering barrier would also be an "unlog-bit" barrier but it's a different implementation.

I chose the name MMTkUnlogBitBarrierSet because the mmtk-core simply calls the metadata "unlog bit" or just "log bit".  But I think it is worth emphasizing that it is "object unlog bit" or "object remembering barrier" rather than "field unlog bit" or "field remembering barrier".

But I hesitate to change the name now because  "MMTkObjectUnlogBitBarrier" and "MMTkObjectRememberingBarrier" are both similar to the current "MMTkObjectBarrier", and the longer name feels like it should be a subclass rather than superclass.  It is also a bit confusing because both "ObjectBarrier" and "SATBBarrier" are object-logging.  I think we should rename "ObjectBarrier" to a more meaningful name, such as "object modification barrier".  We can discuss this in a group.

wks:
Another failure https://github.com/mmtk/mmtk-openjdk/actions/runs/17941785053/job/51021285329?pr=332 was a "duplicated edges" error in SemiSpace.  It was already reported.  See https://github.com/mmtk/mmtk-openjdk/issues/331

wks:
I labelled this as ready for review.

There is one thing intentionally not done in this PR:

To be consistent with write barriers, `MMTkBarrierSet{Assembler,C1,C2}` should also override `load_at` and provide virtual methods `object_reference_read_{pre_post}` for subclasses to override.  However, we currently don't have read barriers in MMTk core except `SATBBarrier` which only applies read barriers for weak references.  We can't work out a nice interface for read barriers unless we have one in MMTk core, as it may be more complicated than write barriers.  For example,

-   "loaded value barriers" is a post barrier, but it should have access to the loaded value, and can override the value read by the raw memory read.  In other words, it needs to overwrite the result register (Assembler and C1), and/or may need to return a different register/node that holds the final result (C1 and C2)
-   "indirection barriers" may override the slot address to read from, and apply not only to reference fields, but value fields as well.

Because `SATBBarrier` is currently the only barrier (policy) that has a barrier (mechanism) for read accesses, we just let it override `load_at` directly.  If we introduce another barrier, such as the field-logging SATB barrier or RC barrier, we may hoist some code to a shared file or superclasses.

qinsoon (commented):
Can you run benchmarks with `GenImmix` to make sure the changes do not affect performance?

openjdk/barriers/mmtkSATBBarrier.cpp
```diff
@@ -242,156 +108,50 @@ void MMTkSATBBarrierSetC1::load_at_resolved(LIRAccess& access, LIR_Opr result) {
...
+      }
     }
-  } else {
-    __ jump(slow);
   }
+}
 
-  __ branch_destination(slow->continuation());
+void MMTkSATBBarrierSetC1::object_reference_write_pre(LIRAccess& access, LIR_Opr src, LIR_Opr slot, LIR_Opr new_val) const {
+  object_reference_write_pre_or_post(access, src, true);
```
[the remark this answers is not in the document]
wks: I should have comments on those call sites, too.  I'll add them.

I can't have `object_reference_write_pre` and `object_reference_write_post` in `MMTkUnlogBitBarrierSetC1`.  They are virtual methods in `MMTkBarrierSetC1`, and having them in `MMTkUnlogBitBarrierSetC1` will override those virtual methods.  Instead, each concrete barrier type (`MMTkObjectBarrierSetC1` and `MMTkSATBBarrierSetC1`) should override one of them, and not both.  The overriding method, like this `MMTkSATBBarrierSetC1::object_reference_write_pre`, should call a method in `MMTkUnlogBitBarrierSetC1` that has a different name, which is `object_reference_write_pre_or_post` in this case.

wks:
> Can you run benchmarks with `GenImmix` to make sure the changes do not affect performance?

Here are the results.  vole.moma, lusearch from DaCapo Chopin, 40 invocations, 5 iterations, with three heap sizes: 1.9x, 2.4x and 3.0x w.r.t. SemiSpace on vole.moma.  build1 is master and build2 is this PR.

https://squirrel.anu.edu.au/plotty/wks/noproject/#0|vole-2025-09-25-Thu-081657&build^hfac^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&hfac^mmtk_gc&build;build1|40&Histogram%20(with%20CI)^build^mmtk_gc&

Most data show that there is no difference.  But some cases, such as `(build1, 1.8x, Immix)` and `(build2, 2.4x, StickyImmix)` exhibited large slowdown together with large noise.  They are caused by abnormal slowdowns in *mutator time* (time.other) in one or two invocations.  This can be seen by plotting the time of each invocation separately.

-   `(build1, 1.8x, Immix)`: https://squirrel.anu.edu.au/plotty/wks/noproject/#0|vole-2025-09-25-Thu-081657&build^hfac^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4&mmtk_gc^1^Immix&hfac^1^1892|40&XY^build^invocation&time.other
-   `(build2, 2.4x, StickyImmix)`: https://squirrel.anu.edu.au/plotty/wks/noproject/#0|vole-2025-09-25-Thu-081657&build^hfac^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4&mmtk_gc^1^StickyImmix&hfac^1^2428|40&XY^build^invocation&time.other
-   `(build2, 3.0x, Immix)`: https://squirrel.anu.edu.au/plotty/wks/noproject/#0|vole-2025-09-25-Thu-081657&build^hfac^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4&mmtk_gc^1^Immix&hfac^1^3023|40&XY^build^invocation&time.other

Since it affects both the master branch and this PR, I suspect it is caused by either

-   an existing bug in the master branch that caused abnormal slowdown in mutator, or
-   some anomalies on vole.moma that interrupted the testing processes at unpredictable times.

I'll run the experiment on a different moma machine to see if can be reproduced.

wks:
The CPU frequency of vole.mama, the machine I ran tests on, was not locked to maximum.  This may be the cause of the abnormal slowdown.  I am running the same tests on another machine where CPU frequencies are set to maximum.

qinsoon:
> > Can you run benchmarks with `GenImmix` to make sure the changes do not affect performance?
> 
> Here are the results. vole.moma, lusearch from DaCapo Chopin, 40 invocations, 5 iterations, with three heap sizes: 1.9x, 2.4x and 3.0x w.r.t. SemiSpace on vole.moma. build1 is master and build2 is this PR.
> 
> [https://squirrel.anu.edu.au/plotty/wks/noproject/#0|vole-2025-09-25-Thu-081657&build^hfac^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&hfac^mmtk_gc&build;build1|40&Histogram%20(with%20CI)^build^mmtk_gc&](https://squirrel.anu.edu.au/plotty/wks/noproject/#0%7Cvole-2025-09-25-Thu-081657&build%5Ehfac%5Einvocation%5Eiteration%5Emmtk_gc&GC%5Etime%5Etime.other%5Etime.stw&%7C10&iteration%5E1%5E4%7C20&1%5Einvocation%7C30&1&hfac%5Emmtk_gc&build;build1%7C40&Histogram%20(with%20CI)%5Ebuild%5Emmtk_gc&)

Can you run *all* the benchmarks over the weekend?

tianleq:
> Results for all benchmarks are here. DaCapo Chopin MR2, vole.moma, 2.4x min heap size w.r.t semispace on vole.moma, 40 invocations, 5 iterations, using ConcurrentImmix, Immix and GenImmix.
> 
> [https://squirrel.anu.edu.au/plotty/wks/noproject/#0|vole-2025-09-26-Fri-092914&benchmark^build^hfac^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^hfac^mmtk_gc&build;build1|60&build&mmtk_gc|70&build-mmtk_gc^build-gc|40&Histogram%20(with%20CI)^build-mmtk_gc^benchmark&](https://squirrel.anu.edu.au/plotty/wks/noproject/#0%7Cvole-2025-09-26-Fri-092914&benchmark%5Ebuild%5Ehfac%5Einvocation%5Eiteration%5Emmtk_gc&GC%5Etime%5Etime.other%5Etime.stw&%7C10&iteration%5E1%5E4%7C20&1%5Einvocation%7C30&1&benchmark%5Ehfac%5Emmtk_gc&build;build1%7C60&build&mmtk_gc%7C70&build-mmtk_gc%5Ebuild-gc%7C40&Histogram%20(with%20CI)%5Ebuild-mmtk_gc%5Ebenchmark&)
> 
> time.stw is a bit noisy, but overall there is no clear sign of regression or improvement.
> 
> time.other shows no obvious changes, too. There is one noisy benchmark `sunflow` which shows 15% speed-up in mutator time in GenImmix. Plotting all invocations (as shown below), we don't see any spikes in the results, but GenImmix seems to be constantly faster on build2 (this PR) than build1 (master).
> 
> [https://squirrel.anu.edu.au/plotty/wks/noproject/#0|vole-2025-09-26-Fri-092914&benchmark^build^hfac^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4&benchmark^1^sunflow|60&build&mmtk_gc|70&build-mmtk_gc^build-gc|41&XY^build-mmtk_gc^invocation&](https://squirrel.anu.edu.au/plotty/wks/noproject/#0%7Cvole-2025-09-26-Fri-092914&benchmark%5Ebuild%5Ehfac%5Einvocation%5Eiteration%5Emmtk_gc&GC%5Etime%5Etime.other%5Etime.stw&%7C10&iteration%5E1%5E4&benchmark%5E1%5Esunflow%7C60&build&mmtk_gc%7C70&build-mmtk_gc%5Ebuild-gc%7C41&XY%5Ebuild-mmtk_gc%5Einvocation&)
> 
> From the un-normalized plots shown below, we see build2-GenImmix is even faster than Immix in mutator time.
> 
> [https://squirrel.anu.edu.au/plotty/wks/noproject/#0|vole-2025-09-26-Fri-092914&benchmark^build^hfac^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|60&build&mmtk_gc|70&build-mmtk_gc^build-gc|40&Histogram%20(with%20CI)^build-mmtk_gc^benchmark&](https://squirrel.anu.edu.au/plotty/wks/noproject/#0%7Cvole-2025-09-26-Fri-092914&benchmark%5Ebuild%5Ehfac%5Einvocation%5Eiteration%5Emmtk_gc&GC%5Etime%5Etime.other%5Etime.stw&%7C10&iteration%5E1%5E4%7C20&1%5Einvocation%7C60&build&mmtk_gc%7C70&build-mmtk_gc%5Ebuild-gc%7C40&Histogram%20(with%20CI)%5Ebuild-mmtk_gc%5Ebenchmark&)
> 
> If we assume C2-compiled code is dominating the execution time, then the only change I made to the C2 barrier is removing the `slot` and `new_val` arguments in the slow-path call. But if write barrier slow paths dominate the execution time of sunflow so much that two registers may affect the performance, then build1-GenImmix should be significantly slower than Immix.
> 
> `perf` didn't show any significant bottleneck in build1-GenImmix.

If `slot` gets removed, does this still work with filed barriers? I noticed that code patching also gets removed, I presume it is because slot is removed and thus nothing needs to be patched.

wks:
> If `slot` gets removed, does this still work with filed barriers?

No.  It won't work with field barriers.  That's why the methods are implemented in `MMTkUnlogBitBarrierSet{Assembler,C1,C2}` instead of `MMTkBarrierSet{Assembler,C1,C2}`.  To use field barriers, we need to reimplement a new write barrier fast path anyway.  If the fast path fails, it will need to call the slow path with the `slot` argument.  We will do that in something like `MMTkFieldUnlogBarrierSet{Assembler,C1,C2}`.

The code blobs for calling runtime functions, i.e. `MMTkBarrierSetC1::_object_reference_write_{pre,post,slow}_c1_runtime_code_blob` still properly forward the `src`, `slot` and `new_val` arguments to the runtime functions as long as those arguments are provided, and should still work for field barriers.

> I noticed that code patching also gets removed, I presume it is because slot is removed and thus nothing needs to be patched.

That's right.  If we don't need the slot address, we don't need to code patching.  We will need code patching again to support field barriers.

qinsoon:
There are 3 tests failing in concurrent Immix.

wks:
I fixed one obvious bug which is a typo that removes the barrier when the new field value is null.  https://github.com/mmtk/mmtk-openjdk/pull/332/commits/940c6915a8d39661cb78112afc97ef4b80cb1305

But the SIGSEGV bug remains.  It is not related to the weak ref load barrier because it can still be reproduced after disabling reference processing and the ref-load barrier.  When I disable the C1 write barrier fast path, the bug can't be reproduced.  I think it is the culprit.  It is the least favorite part of the refactoring because it tries to work around several limitations of the C1 IR.  I'll look into that later.

tianleq:
> I fixed one obvious bug which is a typo that removes the barrier when the new field value is null. [940c691](https://github.com/mmtk/mmtk-openjdk/commit/940c6915a8d39661cb78112afc97ef4b80cb1305)
> 
> But the SIGSEGV bug remains. It is not related to the weak ref load barrier because it can still be reproduced after disabling reference processing and the ref-load barrier. When I disable the C1 write barrier fast path, the bug can't be reproduced. I think it is the culprit. It is the least favorite part of the refactoring because it tries to work around several limitations of the C1 IR. I'll look into that later.

I ran into the same issue when I was implementing the prototype. 

I also ran into some C1 issues and it might be related. So when code needs patching, the value of the slot is invalid. The comment in the following is incorrect, after doing those the scratch register still contains an invalid slot. The only reason of doing this is to make the c1 LIR assembler happy, otherwise, it might trigger an assertion failure saying something like "expecting a register but found an address"

```
if (stub->patch_code != lir_patch_none) {
    // Patch
    assert(stub->scratch->is_single_cpu(), "must be");
    assert(stub->scratch->is_register(), "Precondition.");
    ce->mem2reg(stub->slot, stub->scratch, T_OBJECT, stub->patch_code, stub->info, false /*wide*/, false /*unaligned*/);
    // Now stub->scratch contains the pre_val instead of the slot address
    // So the following is to load the slot address into scrach register
    // Resolve address 
    auto masm = ce->masm();
    LIR_Address* addr = stub->slot->as_address_ptr();
    Address from_addr = ce->as_Address(addr);
    __ lea(stub->scratch->as_register(), from_addr);
    // Store parameter
    ce->store_parameter(stub->scratch->as_pointer_register(), 1);
  }

```

openjdk/barriers/mmtkUnlogBitBarrier.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+/// This file contains abstract barrier sets for barriers based on the (object-grained) unlog bit.
+
+struct MMTkC1UnlogBitBarrierSlowPathStub;
+
+const uintptr_t UNLOG_BIT_BASE_ADDRESS = GLOBAL_SIDE_METADATA_VM_BASE_ADDRESS;
+
+//////////////////// Runtime ////////////////////
+
+class MMTkUnlogBitBarrierSetRuntime: public MMTkBarrierSetRuntime {
```
qinsoon: We never use the term 'unlog bit barrier' anywhere. If you just need a name for the parent class of both SATB barrier and object barrier, you can use something like 'GeneralObjectBarrier' or 'AbstractObjectBarrier'. if you think 'unlog bit barrier' is better, it is okay with me.

qinsoon (approved):
LGTM

openjdk/barriers/mmtkUnlogBitBarrier.hpp
wks: Let's call it UnlogBitBarrier for now because it's what this abstraction is designed for, i.e. barriers that use the unlog bits.  The term "object barrier" is a bit ambiguous for now because the current `ObjectBarrier` is actually a generational remembered set barrier.  We can discuss in a meeting and choose a better name.  And according to https://github.com/mmtk/mmtk-core/issues/1406, we may have a class called "GenerationalRemSetBarrier" that uses both object-remembering and field-remembering (for example, use object-remembering for scalars, and field-remembering for arrays).  I intended to make `MMTkUnlogBitBarrierSetRuntime` a "mix-in".  In C++'s terminology, it can be one field-less no-virtual super-class in a class that inherits from multiple super-classes.  In that case, a `GenerationalRemSetBarrier` that inherits from both `UnlogBitBarrier` and `FieldUnlogBitBarrier`, where the latter will be a "mix-in" that handles a separate field-logging bits metadata.

PR #336, opened 2025-12-01
Demote re-registered nmethods

no-defun-allowed:
If an nmethod is in the mature roots and the nmethod is re-registered due to code patching, the nmethod will appear in both the nursery roots and the mature roots for one collection, causing some roots to be scanned twice. This PR removes an nmethod from the mature roots, when the nmethod is re-registered, so that the nmethod will only be present in the nursery.

### DIFF
--- a/mmtk/src/api.rs
+++ b/mmtk/src/api.rs
@@ -512,6 +512,13 @@ pub extern "C" fn mmtk_register_nmethod(nm: Address) {
     NMETHOD_SLOTS.with_borrow_mut(|slots| {
         if !slots.is_empty() {
             let mut roots = crate::NURSERY_CODE_CACHE_ROOTS.lock().unwrap();
+            let mut mature_roots = crate::MATURE_CODE_CACHE_ROOTS.lock().unwrap();
+            // The nmethod might already be in the mature roots, if we
+            // are re-registering the nmethod due to code patching.
+            // If the nmethod is already in the mature roots, we demote
+            // the nmethod to the nursery roots, as the nmethod might
+            // now refer to a young object.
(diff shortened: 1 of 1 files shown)


k-sareen:
Closes https://github.com/mmtk/mmtk-openjdk/issues/331

wks:
Could you ~~re-open this PR to~~ target the `jdk-11` branch, instead?  You can open another PR for `jdk-21` if you want to.

*(Edit: It turned out that there is a way to set the base branch without re-opening the PR.  I have made the change already.)*

mmtk/src/api.rs
```diff
@@ -512,6 +512,13 @@ pub extern "C" fn mmtk_register_nmethod(nm: Address) {
     NMETHOD_SLOTS.with_borrow_mut(|slots| {
         if !slots.is_empty() {
             let mut roots = crate::NURSERY_CODE_CACHE_ROOTS.lock().unwrap();
+            let mut mature_roots = crate::MATURE_CODE_CACHE_ROOTS.lock().unwrap();
```
wks, line 515: Holding two locks at the same time is prone to dead-locking, especially when `ScanCodeCacheRoots::do_work` acquires the two locks in the opposite order.  One way to fix it is putting the two hash maps under the same lock.  I refactored this part of code locally when trying to address the "fix relocation" thing, but haven't merged it yet because it is somewhat unrelated to the "fix relocation" itself.  I'll make another pull request later for the refactoring.

wks (approved):
LGTM

Let's merge this obvious fix for now, and I'll refactor the mutex later, as I mentioned in the other comment.

PR #337, opened 2025-12-03
Demote re-registered nmethods

wks:
If an nmethod is in the mature roots and the nmethod is re-registered due to code patching, the nmethod will appear in both the nursery roots and the mature roots for one collection, causing some roots to be scanned twice. This PR removes an nmethod from the mature roots, when the nmethod is re-registered, so that the nmethod will only be present in the nursery.

This PR is ported from https://github.com/mmtk/mmtk-openjdk/pull/336 for the `jdk-21` branch.


wks:
@no-defun-allowed Benchmarks show that it is still possible for OpenJDK 21 to re-registers nmethods.  So this change is still needed for OpenJDK 21.

qinsoon (approved):

PR #347, opened 2026-03-18
Fix signedness error of max LOS size in C2

wks:
The `PlanConstraints::max_non_los_default_alloc_bytes` constant is a `usize` and can be `usize::MAX` for some plans including NoGC.  The C2 code erroneously considered it signed and did signed comparison against object sizes.  Consequently, when allocating dynamically sized arrays, the C2-generated code always goes to the slow path.  It slows down certain workloads (such as hsqldb and luindex in DaCapo 2006) by 20% in total time.

We fix this bug by always performing 64-bit unsigned comparison.

### DIFF
--- a/mmtk/Cargo.lock
+++ b/mmtk/Cargo.lock
@@ -28,9 +28,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.13"
+version = "1.0.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


wks:
Several DaCapo tests failed during "Fetch dacapo-23.9-RC3-chopin cache".  It looks like some transient network error.

wks:
We have used Mergify to backport the PR for jdk-21 to jdk-11.

https://github.com/mmtk/mmtk-openjdk/pull/350

PR #348, opened 2026-03-18
Fix signedness error of max LOS size in C2

wks:
The `PlanConstraints::max_non_los_default_alloc_bytes` constant is a `usize` and can be `usize::MAX` for some plans including NoGC.  The C2 code erroneously considered it signed and did signed comparison against object sizes.  Consequently, when allocating dynamically sized arrays, the C2-generated code always goes to the slow path.  It slows down certain workloads (such as hsqldb and luindex in DaCapo 2006) by 20% in total time.

We fix this bug by always performing 64-bit unsigned comparison.

### DIFF
--- a/mmtk/Cargo.lock
+++ b/mmtk/Cargo.lock
@@ -28,9 +28,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.13"
+version = "1.0.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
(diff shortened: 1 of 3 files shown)


qinsoon (approved):
LGTM

qinsoon:
I opened a PR to allow easier backporting to `jdk-11`. We should consider merging that PR first, and try the workflow with this PR.

qinsoon:
https://github.com/mmtk/mmtk-core/pull/1456 changed the default `max_non_los_default_alloc_bytes` from signed int32 max to unsigned uint64 max, which revealed this bug. It is just incidental that we only observed slowdown for NoGC from this issue, as every other plan overrides the value `max_non_los_default_alloc_bytes`.

PR #349, opened 2026-03-22
Add configurations for mergify to allow backporting to jdk-11

qinsoon:
Copied from mmtk-julia's configuration: https://github.com/mmtk/mmtk-julia/blob/master/.github/mergify.yml. 

When a PR that is based on `jdk-21` with a label `backport-jdk-11` is merged, a PR based on `jdk-11` will be created by cherry-picking the changes.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,12 @@
+pull_request_rules:
+  - name: backport to jdk-11
+    conditions:
+      - base=jdk-21
+      - label=backport-jdk-11
+    actions:
+      backport:
+        branches:
+          - jdk-11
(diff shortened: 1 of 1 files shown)


wks (approved):
LGTM

wks:
The direction of porting is OK.  OpenJDK 21 is easier to work with on modern Linux distributions with newer GCC and libstdc++ versions, while OpenJDK 11 has problem compiling on distros that ship the newest GCC, such as ArchLinux.  My usual work flow is developing on jdk-21, making sure it compiles and works, and then cherry-picking it to jdk-11.

PR #351, opened 2026-03-26
Improve pgo-build.sh

qinsoon:
1. Reduce hard-coded path/version/etc, making it easier to maintain and use in both jdk11 and jdk21.
2. Add a step to configure jdk, so the script now works on fresh cloned repos.


wks (approved):
LGTM

PR #359, opened 2026-07-10
Add field-logging barrier and field unlog bit (JDK21)

wks:
Upstream PR: https://github.com/mmtk/mmtk-core/pull/1528


mmtk/src/slots.rs
```diff
@@ -205,6 +205,11 @@ impl<const COMPRESSED: bool> Slot for OpenJDKSlot<COMPRESSED> {
             unsafe { self.addr.store(object) }
         }
     }
+
+    fn to_address(&self) -> Address {
+        assert!(!COMPRESSED);
+        self.addr
```
k-sareen, line 211: I just realized this issue in my branch. I think something like the below is better:

```suggestion
    if COMPRESSED {
        self.untagged_address()
    } else {
        self.addr
    }
```

wks: I will eventually do this.  Currently the 1 bit per word field unlog bit side metadata doesn't work for compressed pointers, so I want it to fail loud.  To support compressed pointers, I will introduce an alternative 1 bit per 32 bit field unlog bit side metadata and give the VM binding the option to select that.

k-sareen: Ah fair enough. Makes sense. I just found this bug in my branch so I thought it'd be better to make a comment here in case you also have issues with the same in the future.

PR #364, opened 2026-08-05
Use actions/cache@v4. Set max parallel to 10

qinsoon:
We saw intermittent errors when restoring dacapo cache in our CI. Those jobs failed at the step of 'Fetch dacapo-23.11-MR2-chopin cache' with half downloaded dacapo from cache. 

This PR updates the cache action to v4, and limit the max parallel job to 10 for the testing. Instead of having 34 jobs downloading the 6GB dacapo jar from cache, we will only have 10 concurrent jobs downloading from cache. Hopefully this will improve the stability.

### DIFF
--- a/.github/workflows/run-dacapo-chopin-inner.yml
+++ b/.github/workflows/run-dacapo-chopin-inner.yml
@@ -23,7 +23,7 @@ jobs:
     steps:
       - name: Check ${{ env.DACAPO_VERSION }} cache
         id: check-cache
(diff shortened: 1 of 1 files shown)


qinsoon:
Still see fails for downloading the dacapo cache. I will see if we can add a retry for the step.

qinsoon:
The issue was not about downloading cache. The runner ran out of disk during downloading the cache.

PR #367, opened 2026-08-10
Swap the maximize-build-space action to free-disk-space

qinsoon:
We frequently see some jobs hung while fetching Dacapo cache. The error from the runner indicated that this is due to out-of-space.

We use `maximize-build-space` to get more disk space for Github runner. But it seems the action is not very effective (https://github.com/easimon/maximize-build-space/issues/44), and may even reduce the available disk space (https://github.com/easimon/maximize-build-space/issues/55). This PR replaces `maximize-build-space` with a different action `free-disk-space`. It seems no out-of-disk in the current runs.

### DIFF
--- a/.github/workflows/run-dacapo-chopin-inner.yml
+++ b/.github/workflows/run-dacapo-chopin-inner.yml
@@ -70,17 +70,16 @@ jobs:
     steps:
       - name: Check free space
         run: df -h
-      - name: Maximize build space
-        uses: easimon/maximize-build-space@master
+      - name: Free disk space
+        uses: jlumbroso/free-disk-space@v1.3.1
         with:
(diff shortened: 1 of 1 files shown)


k-sareen:
We can also ask @steveblackburn to provide a smaller DaCapo JAR. The current one includes the `large` and `vlarge` sizes which are unnecessary for CI testing.

qinsoon:
> We can also ask @steveblackburn to provide a smaller DaCapo JAR. The current one includes the `large` and `vlarge` sizes which are unnecessary for CI testing.

I can do this in a separate PR.

k-sareen (approved):
LGTM. Thanks!

PR #369, opened 2026-08-10
Use the Dacapo Chopin minimal zip for CI

qinsoon:
As @k-sareen suggsted in https://github.com/mmtk/mmtk-openjdk/pull/367#issuecomment-5235111735, we can use a minimal dacapo chopin zip in our CI (1.4G), instead of the full zip (6.3G).


k-sareen (approved):
LGTM. Thanks!

PR #372, opened 2026-08-14
No longer report pmd digest verification fail

qinsoon:
Related to https://github.com/mmtk/mmtk-openjdk/issues/262. Not showing the benchmark as fail in our CI, if we see the same error.

Done by Claude.

### DIFF
--- a/.github/scripts/ci-matrix-result-check.py
+++ b/.github/scripts/ci-matrix-result-check.py
@@ -95,7 +95,7 @@ def read_in_expected_results(build, benchmark):
 
     return data["results"][arch][build][benchmark]
 
-def print_log(directory, search_string):
+def read_log(directory, search_string):
     import gzip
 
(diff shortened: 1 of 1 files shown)


k-sareen (commented):
I'm a bit scared that this will hide real issues. Graal seems to skip this error code as well but I kinda don't understand how/where the function is called in Graal's CI.

https://github.com/oracle/graal/commit/8f576fa8e353f07d2642ef06ce2c914b6ad7de14

I think perhaps we should discuss this in a meeting before we merge this PR.

k-sareen:
I see. I think Graal just does `--no-validation` specifically for `pmd`: https://github.com/oracle/graal/commit/531528724b71131a3fadfa10288ccb8c28bc6482

qinsoon:
See https://github.com/mmtk/mmtk-openjdk/pull/373.

PR #373, opened 2026-08-14
Use --no-validation for dacapo when running pmd

qinsoon:
Related to https://github.com/mmtk/mmtk-openjdk/issues/262. Skip digest validation for pmd.

Done by Claude.


k-sareen:
I think it's probably fine now, but it'd still be good to discuss in meeting before we merge it, if that's alright with you.

qinsoon:
> I think it's probably fine now, but it'd still be good to discuss in meeting before we merge it, if that's alright with you.

Sure. It is not urgent. Let's discuss this in a team meeting before merging it.

